PR #220, opened 2012-03-28
Pass final redirect url to callback

ashishgandhi:
Useful if you need to know the final URL where the redirect ends. Use case for this is when you have logic/rules based on domain name of a page.


bradvogel:
@mikeal Also curious about above comment

PR #680, opened 2013-10-10
Socks5 support

### LINKED ISSUE #141: Add Socks Proxy support
Cause there is no hope that this will be supported by nodejs itself (https://github.com/joyent/node/issues/1973) I hope this library can support socks proxies. A simply example where it is useful is the tor network, which only supports the socks protocol.

mattcg:
This is a replacement for #648 and fixes #141.

The patch adds [mattcg/socks5-http-client](https://github.com/mattcg/socks5-http-client) and [mattcg/socks5-https-client](https://github.com/mattcg/socks5-https-client) as dependencies so that request can use the agent objects from both to support SOCKS proxying.

A test is included. To run it you should have a SOCKS5 server running on localhost:1080. Use the following command to start one if you have sshd running locally:

`ssh -N -D 0.0.0.0:1080 localhost`

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+tests
+node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: node_js
+node_js:
+  - 0.8
+  - 0.10
+
+env:
+  - OPTIONALS=Y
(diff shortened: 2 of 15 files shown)


raitucarp:
socks proxy by mattcg is break on node 0.12

Issue #794, opened 2014-02-03
Error: Cookie not in this host's domain.


breezewish:
Such kind of tough-cookie errors should be ignored, however currently they cause errors and break normal requesting.

For example, assume `http://foo.bar.com` returns headers like `Set-Cookie: NAME=value; path=/; domain=hello.bar.com`. It is an invalid cookie.
With the newest `request` module, this request will fail. I do not think this is a correct behaviour. These errors should be ignored.

jtara1:
what would be the default domain if domain wasn't specified in the response cookie string
```
> jar.setCookie(req.cookie('abcaaa=aaa'))
Cookie="abcaaa=aaa; Path=/; hostOnly=true; aAge=0ms; cAge=0ms"
> jar.getCookieString()
''
```

PR #801, opened 2014-02-06
Ignore cookie parsing and domain errors

lalitkapoor:
this addresses #794

/cc @stash @mikeal


apsavin:
+1 for change the default behavior (we should not ignore errors by default)
We need to provide an option to do not ignore errors [here](https://github.com/lalitkapoor/request/commit/7fb164731a5aad80c6539e33eda4ad4a51bb7871) at least.

PR #884, opened 2014-04-29
Documented tough-cookie installation.

wbyoung:
I was a little confused about using cookie jars. I got it working, but am not sure if this is the correct way or not. For me, in order for it to work, `though-cookie` needs to be installed. If this is how it's supposed to be, this pull request updates the docs to point that out.


LoicMahieu:
I merged this PR but it was a mistake. `tough-cookie` is builtin and required by `request` for a while. There is no need for user to install it manually. So I revert the PR. Sorry for confusion.

PR #1064, opened 2014-09-10
Fixing HTTP redirects "not carrying pre-defined host-header value when the Location header is not a URL".

umutm:
Fixing HTTP redirects "not carrying pre-defined host-header value when the Location header is not a URL".


elyran:
So what was the solution? I'm trying to override the DNS lookup phase in a `request()` call.

elyran:
These days HTTPS is must. The ref is about HTTP... Any solution?

PR #1132, opened 2014-10-06
Correct infinite loop with the same location

blakeembrey:
The issue is apparent when loading from Google Drive (e.g. `https://dd7777360edd06ed6fa7cbc0725dbd736cbda015.googledrive.com/host/0B9Vee7pZ81xmUzBqUmluT3FWVkU/my.js`). The root cause is that the `304 Not Modified` responds with a location header, which results in an infinite loop to the same url. This patch only does a redirect now if the url has changed.

### DIFF
--- a/request.js
+++ b/request.js
@@ -883,6 +883,10 @@ Request.prototype.onResponse = function (response) {
     var location = response.caseless.get('location')
     debug('redirect', location)
 
+    if (!isUrl.test(location)) {
+      location = url.resolve(self.uri.href, location)
+    }
+
     if (self.followAllRedirects) {
       redirectTo = location
     } else if (self.followRedirect) {
@@ -973,7 +977,7 @@ Request.prototype.onResponse = function (response) {
     }
   }
 
-  if (redirectTo && self.allowRedirect.call(self, response)) {
+  if (redirectTo && redirectTo !== self.uri.href && self.allowRedirect.call(self, response)) {
     debug('redirect to', redirectTo)
 
     // ignore any potential response body.  it cannot possibly be useful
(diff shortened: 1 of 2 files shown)


nylen:
I realized that we keep sending the same `If-Modified-Since` header after a redirect, so of course the server is going to send the same redirect again!

This works:

```js
var request = require('request');

require('request-debug')(request);

var u = 'https://dd7777360edd06ed6fa7cbc0725dbd736cbda015.googledrive.com/host/0B9Vee7pZ81xmUzBqUmluT3FWVkU/my.js';

request(u, {
    headers : {
        'if-modified-since' : 'Mon, 06 Oct 2015 06:43:27 GMT'
    },
    followRedirect : function(res) {
        return true;
        // return (res.statusCode !== 304);
    }
}, function(err, res, body) {
    console.log(err, res && res.statusCode, body);
}).on('redirect', function() {
    console.log('redirect', this.response.statusCode, this.uri.href, this.getHeader('if-modified-since'))
    this.removeHeader('if-modified-since')
});
```

saper:
I think wget's behaviour is better.

3xx family of reponse codes belong to the ["Redirection family" per RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.1), so maybe they set `Location` for all 3xx responses (just guessing). 

````
The first digit of the Status-Code defines the class of response. The last two digits do not have any categorization role. There are 5 values for the first digit:

(...)
      - 3xx: Redirection - Further action must be taken in order to
        complete the request
````

304 handling should be done first, before looking at the `Location` header. Only if we know don't have it cached, we should follow up with a non-conditional GET, in that case we should use the `Location` header, whatever location it points to.

Issue #1249, opened 2014-11-07
maxSockets does not limit the number of socket connections


anklos:
Hi folks,

I am trying to use `pool({maxSockets: 100})` to increase the default `maxSockets` number. I find that once I set `pool` in the option, no matter what value you give to `maxSocket`, it seems always enable maxSockets to be `Infinity`.

My local test:

``` coffee
## index.coffee
restify = require 'restify'
client  = require './client'

server = restify.createServer()

server.get '/test', client.send

server.listen 8080, () -> console.log 'server started'

```

``` coffee
## client.coffee

request = require 'request'

exports.send = (req, res, next) ->
  options =
    url: "https://google.com.au"
    strictSSL: true
    secureProtocol: true
    pool:
       maxSockets: 5

  request options, (err, response, body) ->
    res.send 200, body
```

so if you `npm start` and go to `localhost:8080/test`, you would send a get request to `https://google.com.au`

Now I use the performance tool `wrk` to run this test: `wrk -t12 -c3000 -d30s http://127.0.0.1:8080/test`, which initiates 12 threads to create 3000 connections and last 30 seconds.

Next get the `ip` of google.com.au by `host google.com.au` and I get:

```
google.com.au has address 74.125.237.111
google.com.au has address 74.125.237.127
google.com.au has address 74.125.237.120
google.com.au has address 74.125.237.119
```

Then in the console I do  `watch "netstat -an | grep 74.125 | wc -l`

From the output I can see there are always hundreds of socket connections opening  no matter what value I give `maxSockets`. But If I don't set `pool : { maxSockets: xxx}` in the `options`, I will only see around 10 opening sockets from console.

Is my way of using `maxSockets` correct? Do I miss something here?

Thanks

Issue #1150, opened 2014-10-13
How to set maxSockets for all requets to be


ekelvin:
I tried the request.defaults to set hte maxSocket: 3 but the problem still persists, thousands of requests are sent to the server in parallel :S :(. 
here is my try, notice download is with callback

```
request.defaults({ pool: { maxSockets: 3 } });
var download = function (uri, filename, callback) {        
        request.head(uri, function (err, res, body) {
            if (err) {
                console.error(err);
                return;
            }                    
            if (res.headers['content-type'] == "image/jpeg") {
                request(uri).pipe(fs.createWriteStream(filename)).on('close', callback);
            }
            else
                console.error('content-type:' + res.headers['content-type']);
        });        
};
```

PR #1251, opened 2014-11-07
Clarify that defaults() does not modify global defaults

nylen:
This has come up several times recently, for example #1150 and #1249.


fritx:
What if I need to global setting some options, such as strictSSL: false?

Issue #1268, opened 2014-11-14
Can I detect the statusCode of the request and then pipe it to a stream?


tybenz:
I want to pipe the response to a request to the reponse of my own server. But I want to render my own error page if the status code comes back anything but 200.

``` javascript
app.get( '/route', function( req, res, next ) {
    request( 'https://service.com/endpoint', function( err, response ) {
        if ( err || response.statusCode != 200 ) {
            sendError();
        }
    }).pipe( res );
});
```

This is not working for me. I think I'm having issues understanding the order of events/how stream actually work. Is this at all possible?

tybenz:
This works well for me:
```javascript
app.get('/route', function(req, res, next) {
  request('http://domain.com')
    .on('response', function(response) {
       if (response.statusCode === 200)
          return;
       this.abort();
       next(new Error('there was something wrong with this request'));
    })
    .pipe(res)
  ;
});
```
But this doesn't:
```javascript
app.get('/route', function(req, res, next) {
  request('http://domain.com')
    .on('response', function(response) {
       if (response.statusCode === 200)
          return;
       this.abort();
      render(); // where controller.render calls res.send eventually but there is a delay.
    })
    .pipe(res)
  ;
});
```

I get the following error message: `Error: Can't set headers after they are sent.` With the second approach. If I call `res.send( 'foobar' )` instead of `render`, it works.

`render` ends up calling `res.send`, but it takes time to get there. I'm thinking that part of the pipe actually happens and the headers for `res` are set, and then I try to call `res.send` again later.

How come `this.abort` isn't wiping `res` clean again?

PR #1270, opened 2014-11-15
Res stream example

tbuchok:
this build off the previous PR https://github.com/request/request/pull/1269

> merging in order should be fine, i made this difficult didn't i? ... sigh, time for the weekend :boom:

* adds additional example for handling "response" events, copied from https://github.com/request/request/issues/1268


FredKSchott:
Hey @tbuchok, sorry for leaving you hanging with this PR. I'm not sure if you're still interested in pushing this through, but I'd love to get it in. If you still are, here is what I would need to merge:

**Each section/example should have a single purpose, usually to introduce 1 concept.**

So for a simple example of what I mean: your first example introduces the idea of listening for the `response` event, which is great. But the example is titled **Responses: `statusCode`, `headers`, et. al** which sounds like it is introducing lots of concepts. This could easily overwhelm a newbie, which is bad. 

Another example: Your entire second code block is very overwhelming, even for me. What is your purpose for including it (what 1 concept do you want someone to learn from reading it) and whats the fewest number of lines you can convey that concept in?

tbuchok:
agreed it's overwhelming. let's close and re-eval. there's mega discussions on docs right now, let's see where that lands and we can wiggle a couple smaller examples into place when ready. thanks again for the review, fred.

FredKSchott:
@tbuchok Understood, thanks for taking the time!

PR #1277, opened 2014-11-19
Add Coveralls configuration

simov:
@nylen you just need to enable coveralls for the request project on https://coveralls.io, other than that technically it should work. I tried it on one of my projects with _mocha_ instead of _tape_ and it seems to be fine.

Also you can try this line out

``` bash
./node_modules/.bin/istanbul cover ./node_modules/tape/bin/tape tests/test-*.js
```

This will generate a nice html output in `coverage/lcov-report/index.html`, so you can check the output locally as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 coverage
+.idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## Change Log
 
+### v2.49.0 (2014/11/28)
+- [#1295](https://github.com/request/request/pull/1295) fix(proxy): no-proxy false positive (@oliamb)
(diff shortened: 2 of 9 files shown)


FredKSchott:
@simov +1 on moving converalls out of the PR comments themselves and into gitter or travis output. I've seen a few complaints of how noisy it is while they never seem to add much value to the PR

simov:
The build just have to fail if the coverage is going down, not sure if that's possible though.

nylen:
Okay, coveralls comments on PRs should be disabled now.  I think getting Coveralls output into somewhere else would be a bit of work.

I've submitted #1343 to add a coverage badge to the readme.  If @coveralls doesn't show up in there then we should be good to go.

nylen:
OK now they're really disabled.

PR #1297, opened 2014-12-02
Cast timeout to a number.

Raynos:
This creates consistency in the timeout feature.

The `setTimeout` implementation in node will correctly cast any strings passed in to a number using `*= 1`

This means that 

```js
request({
  uri: '/foo',
  timeout: '5000'
}, cb);
```

will work as expected.

**HOWEVER** the `req.setTimeout` method does not cast strings to number. It instead defaults to no timeout when passed a string. 

To guard against this bug in node core we manually cast the string to a number in `request/request` itself so that it is consistent.

### DIFF
--- a/request.js
+++ b/request.js
@@ -914,6 +914,8 @@ Request.prototype.start = function () {
   self.req = self.httpModule.request(reqOptions)
 
   if (self.timeout && !self.timeoutTimer) {
+    self.timeout *= 1;
+    
     self.timeoutTimer = setTimeout(function () {
       self.abort()
       var e = new Error('ETIMEDOUT')


FredKSchott:
If you're saying its a problem that Node's `request` object doesn't accept a string value for `request.setTimeout()`, then wouldn't the correct solution be to fix that behavior in node itself? Adding a fix here just moves our behavior farther away from Node's request object without fixing the core problem for all other Node `request` object consumers.

FredKSchott:
I'll motion to close this PR, any objections?

nylen:
+1 to close

PR #1308, opened 2014-12-08
Add browser test to keep track of browserify compability.

eiriksm:
As discussed (or more like suggested) over at issue #1307: A browser test to make sure request can be built with browserify.

Last time it broke it was because one of the dependencies had a dependency which in turn used `fs.readFile`, which obviously does not work well in browsers. But those things are hard to spot unless it is automated.

So to sum it up. This is not for testing that all features actually work in the browser. Just to make sure it will bundle nicely.

One negative thing about this is the added dependencies and the added test build time. In addition to having more tests to "support" of course. For me the positive effects outshines this, but I can understand if someone might make that argument.

Anyway, this is the first take. Open for discussion, of course.

Oh, and thanks once more for making and maintaining this fantastic library. Have a nice day!

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   },
   "scripts": {
     "test": "npm run lint && node node_modules/.bin/taper tests/test-*.js && npm run test-browser",
-    "test-browser": "karma start tests/browser/karma.conf.js",
+    "test-browser": "browserify tests/browser/test.js -o tests/browser/test-browser.js && karma start tests/browser/karma.conf.js",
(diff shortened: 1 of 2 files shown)


nylen:
@eiriksm we're having problems with requests to api.github.com from Travis (see [this build log](https://travis-ci.org/request/request/jobs/46111532#L368) for example).

It's time to change the browser tests over to using a mock server.  I know you can start PhantomJS with the [`--ignore-ssl-errors=true` option](http://phantomjs.org/api/command-line.html) to allow self-signed certs, but I'm not sure about the CORS piece.  Want to help?

eiriksm:
Sure. I'm on it

PR #1327, opened 2014-12-26
Fix errors generating coverage reports.

nylen:
server.close() is async, so treat it as such:

http://nodejs.org/api/net.html#net_server_close_callback

This doesn't seem to matter when test files are run in separate
processes using `npm test`, but it does when they are run in the same
process for the coverage report.

@simov fyi (hopefully this is the only issue with our coverage reporting)

### DIFF
--- a/tests/test-cookies.js
+++ b/tests/test-cookies.js
@@ -91,6 +91,7 @@ tape('custom store', function(t) {
 })
 
 tape('cleanup', function(t) {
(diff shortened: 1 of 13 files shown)


nylen:
Sounds good to me, want to merge this, then I'll submit another PR to change test ports to 6767?  I forgot about this PR so I'm glad you commented.

PR #1329, opened 2014-12-29
Fix memory leak caused by multiple redirect.

Lupino:
When processing multiple redirects Request will init more than once.
the `end event` register every time when the `Request.prototype.start`, also
the same with `pipe event` on `Request.prototype.init`.

you'll see the test scripts here: https://github.com/belltoy/request-redirect-test

### DIFF
--- a/request.js
+++ b/request.js
@@ -956,7 +956,7 @@ Request.prototype.start = function () {
     self.emit('socket', socket)
   })
 
-  if (self._isInitEnd) {
+  if (!self._isInitEnd) {
     self._isInitEnd = true
     self.on('end', function() {
       if ( self.req.connection ) {
--- a/request.js
+++ b/request.js
@@ -651,35 +651,39 @@ Request.prototype.init = function (options) {
     self.agent = self.agent || self.getNewAgent()
   }
 
-  self.on('pipe', function (src) {
-    if (self.ntick && self._started) {
-      throw new Error('You cannot pipe to this stream after the outbound request has started.')
-    }
-    self.src = src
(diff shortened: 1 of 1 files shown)


seanstrom:
@nylen We probably could make your example a private function instead of on the prototype.
Unless we intend to expose that functionality we could just do:

```javascript
function addHandlerOnce(ev, cb) {
  var self = this
  if (!self._addedHandler[ev]) {
    self.on(ev, cb)
    self._addedHandler[ev] = true
  }
}

addHandlerOnce.call(self, ev, cb)
```
or
```javascript
function addHandlerOnce(context, ev, cb) {
  var self = context
  if (!self._addedHandler[ev]) {
    self.on(ev, cb)
    self._addedHandler[ev] = true
  }
}

addHandlerOnce(self, ev, cb)

seanstrom:
Overall I would say that the `init` method should be refactored more.
Not this PR's concern but it would make these changes more maintainable as well as others.

FredKSchott:
@nylen @seanstrom Can we have redirects clear all listeners before calling init again? That way we would consolidate cleanup in one place, instead of scattering it around the module.

@seanstrom +1000 :)

nylen:
That would break any consumers who are listening for those events.  [`request-debug`](https://github.com/request/request-debug) doesn't use these 2 events but it uses a similar approach, and we encourage wrapping request like this.

mxr576:
Any status update about this bug? It is still exists on the latest stable.

FredKSchott:

PR #1330, opened 2014-12-31
Return empty buffer upon empty response body and encoding is set to null

seanstrom:
Instead of returning an empty string, an empty buffer will now be
returned when the encoding is set to null.

This is a PR to finish the work started #1286
Credit to @ramonsnir

@nylen @FredKSchott Review?


FredKSchott:
:+1: Yup looks good!

PR #1331, opened 2015-01-03
Reorders README to be more useful to beginners

rockbot:
Puts the API at the top, with details for how to do more complex things later on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,167 @@ request('http://www.google.com', function (error, response, body) {
 })
 ```
 
+## request(options, callback)
+
+The first argument can be either a `url` or an `options` object. The only required option is `uri`; all others are optional.
+
+* `uri` || `url` - fully qualified uri or a parsed url object from `url.parse()`
+* `qs` - object containing querystring values to be appended to the `uri`
+* `useQuerystring` - If true, use `querystring` to stringify and parse
+  querystrings, otherwise use `qs` (default: `false`).  Set this option to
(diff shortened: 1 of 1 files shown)


simov:
I like it, the most often questions in the chat recently are related to simple options usage.

nylen:
I know our docs could be better organized, but I don't think this is the right way to go.  Our first heading is "Super simple to use", then having all of our 3897 options right underneath that kinda makes me laugh.

How about a link in the first section instead, something like:

> If you need something different then there's probably [an option to customize it](#options).

simov:
Or like big fat 

### [Options](https://github.com/request/request#requestoptions-callback)

link, but you can miss that as well.

Anyway without a fixed sidebar navigation, containing each and every header tag in it, finding info will be only possible through searching using the browser's search.

FredKSchott:
Agreed, "Super simple to use" followed immediately by 100s of options seems glaringly contradictory.

Maybe rename the header for the section on options to address "super simple to use... but super powerful/flexible as well"

simov:
I think _super simple to use_ refers to the API at a whole, not to the options count (although there are some option combinations that are a bit tricky)

nylen:
I still don't like the idea of putting all of our options at the top.  I think it's overwhelming to beginners rather than friendly, and I like the streaming section being next because when I go to the docs usually I am answering the question "ok, how do I do streaming again?"

I think it's time to put some more effort into restructuring our documentation, including splitting it into multiple files because it has gotten really long.  I'm creating a separate issue with some proposed restructuring notes.

PR #1332, opened 2015-01-05
Add clean script to remove test-browser.js after the tests run

seanstrom:
Whenever you run `npm test` it would generate a file.
The location of the file would be `tests/browser/test-browser.js`
which is a directory that the linter will lint.
I tried making the linter ignore the file but it seems we're using
an older version of the linter which may not support ignoring paths.

I may be wrong but this was an alternative solution, safely delete
the file before running the linter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,9 +41,10 @@
     "isstream": "~0.1.1"
   },
   "scripts": {
-    "test": "npm run lint && node node_modules/.bin/taper tests/test-*.js && npm run test-browser",
+    "test": "npm run lint && node node_modules/.bin/taper tests/test-*.js && npm run test-browser && npm run clean",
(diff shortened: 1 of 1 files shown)


simov:
Nice! I have that file hanging around in my repo for quite some time.

nylen:
Why not do this at the end of `test-browser` instead?  That way the file doesn't stick around in our repos.

FredKSchott:
Nice, good catch @seanstrom 
+1 to @nylen, tests should probably be performing cleanup after tests have run

seanstrom:
+1 @nylen I totally didn't think of doing it after the test end. lol

seanstrom:
Would we want a clean-lint script at all?

nylen:
I don't see any need for `clean-lint`.  If you agree, this looks good to me, so can you squash the two commits together and I'll merge?

seanstrom:
Yup sounds good

PR #1336, opened 2015-01-06
Emit error when FormData emits an error event

rwky:
When FormData emits an error event it's not emitted via request, I encountered this when using the Facebook graph API. FormData was emitting an error when the API failed but nothing was catching it.


simov:
That resolved conflict in the last commit looks kind of scary. Although the only problem might be the use of `'` in the code, @rwky , can you start from clean `master` branch and apply your changes on top?

Also would be nice to have a test for that.

nylen:
:+1: to @simov's comments, we want this as a single commit, and it needs tests and a clean build.

What happens now, is an error thrown?  After this change, an error would still be thrown if users of `request` don't handle its `error` event, right?

Do we want to abort the request when this error occurs?

rwky:
Sorry about that the semi-colons were screwing up the test. I've created a less crappy PR #1341

PR #1337, opened 2015-01-06
Standardize test ports on 6767

nylen:
This is a less commonly used port than 8080 so it is less likely to
conflict with other services.  See discussion at #1327.

### DIFF
--- a/tests/test-form-data.js
+++ b/tests/test-form-data.js
@@ -58,7 +58,7 @@ function runTest(t, json) {
       t.ok( data.indexOf('form-data; name="batch"') !== -1 )
       t.ok( data.match(/form-data; name="batch"/g).length === 2 )
 
-      // check for http://localhost:8080/file traces
+      // check for http://localhost:6767/file traces
       t.ok( data.indexOf('Photoshop ICC') !== -1 )
       t.ok( data.indexOf('Content-Type: ' + mime.lookup(remoteFile) ) !== -1 )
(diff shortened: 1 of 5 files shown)


nylen:
where is @coveralls :(

FredKSchott:
:+1: 
Is there any way to set this globally? It's a shame to define it in so many places if we want to keep it standardized

nylen:
For the most part, it is standardized (see `tests/server.js`).  Not all test cases use that to create their servers though.  The best way to fix it would be to port all of the test cases to use it, but that's more effort than I'm willing to undertake now.

We could also add a `tests/vars.js` or similar, but is it really worth adding an extra `require` in all of these files just to store one number?

simov:
On the other hand the unit tests should be as self contained as possible, unless you really have to copy paste huge amount of similar code in each file. I think the port number is not a big problem as it can be fixed with simple find and replace.

FredKSchott:
:+1: Sounds good :)

mikeal:
ideally it could be parametrized so that test could be run in parallel.

nylen:
Sounds tricky.  Switching to the new [Travis docker infrastructure](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/) helped build times a lot.

Are we good to merge this?  I'll do it tomorrow if no one objects or beats me to it.

PR #1342, opened 2015-01-07
Add mock https server and redo start of browser tests for this purpose.

eiriksm:
This is a suggestion ref #1308.

Using a mock https server, and starting karma as a child process. Let me know if I should change something around :)


nylen:
Excellent work, thank you!

eiriksm:
Sounds good :)

And of course it was about time we had tests not relying on 3rd party APIs.

That being said, when trying to get the tests to run with ignoring the self signed cert I noticed that the error message is not so descriptive if you get an "aborted" request like that (meaning the request never really gets a status code) in the browser.

I'll see if I can track it down and/or open an issue about it one of these days

eiriksm:
Ah, sorry. I did not see that PR or take part in that discussion, but I see it now. And makes sense.

If it ends up being defined in one of the test files, it should probably be pulled from there later. As of now (since this is not in master yet) I just changed the port.

eiriksm:
For reference (for me and others), this is the issue in question: #1337

Nice issue number :p

simov:
Actually 1337 is a good port number too :)

nylen:
I want to go ahead and get this in, thanks again @eiriksm!

PR #1344, opened 2015-01-08
Removed need for HTTPS sites to be proxied by tunnels only. Fixes #1293

### LINKED ISSUE #1293: Disable HTTP tunneling for all protocols
From the docs:
- **tunnel** - If `true`, then **always** use a tunneling proxy. If `false` (default), then tunneling will only be used if the   destination is https, or if a previous request in the redirect chain used a tunneling proxy.

What if I need to use a proxy that doesn't support tunneling to make a **https** request?

Shouldn't this module have an option to completely disable tunneling for all protocols?

twolfson:
As discussed in #1293, it is possible to want to do the same HTTP to HTTP proxying when talking to an HTTPS server. This PR allows that by removing the enforcement of HTTPS going to tunneling. In this PR:

- Removed HTTPS requiring usage of HTTP tunnelling
- Added test via @nylen 
- Updated documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ You can still use intermediate proxies, the requests will still follow HTTP forw
 If you specify a `proxy` option, then the request (and any subsequent
 redirects) will be sent via a connection to the proxy server.
 
-If your endpoint is an `https` url, and you are using a proxy, then
+If your endpoint is an `https` url, the `tunnel` option is not `false`, and you are using a proxy, then
 request will send a `CONNECT` request to the proxy server *first*, and
 then use the supplied connection to connect to the endpoint.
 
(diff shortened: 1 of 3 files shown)


nylen:
There was nothing in the test server that was actually responding to that new request.  Here's a working test:  nylen/request@b16edd3

If you can pull that commit in and clean up the docs for the tunnel option (in the [API section](https://github.com/request/request/blob/v2.51.0/README.md#requestoptions-callback)) we'll be good to go.  Thanks for the PR.

twolfson:
Derp. That makes sense, thanks! I will let you know when the update is in.

twolfson:
Alright, I have `cherry-pick`-ed your commit and updated the documentation.

nylen:
Oh... I think I missed something here...

**By default**, we still want to tunnel proxied requests to a HTTPS endpoint.  If we had had a test for that behavior (like we really should have) then that test would have broken with this change.

What we should be adding here is the **option** to disable tunneling HTTPS requests.

Make sense?

twolfson:
Yep, I was considering this a breaking change by not adopting that behavior. I felt like the HTTPS/tunnelling default was a strawman that nobody uses.

I will add in the new logic.

twolfson:
Alright, I have fixed it. For what it's worth, it looks like `request` has been defaulting `undefined` to `tunnel: false`. Techincally, that still makes this a breaking change if people were using HTTPS and `tunnel: false` yet getting tunneling in the past.

twolfson:
I am going to change my implementation of `explicitTunnel` to line up with `explicitMethod` (as in it's a boolean).

twolfson:
Alright, all done and ready for review.

nylen:
The default behavior when requesting a HTTPS endpoint should definitely be to tunnel.  Otherwise the proxy can see the request which is not desired unless you know what you're doing.

I'm working on some more tests to get more comfortable with this change.

twolfson:
Ah, right. Sorry I was being fixated on my use case =/

nylen:
No worries.  It's a good use case, sorry it took me a little while to think through the implications.

It's time to rewrite `tests/test-tunnel.js` to be pure Node.js, using the Node core tests as a guideline:

- [`test/simple/test-http-connect.js`](https://github.com/joyent/node/blob/v0.10.35/test/simple/test-http-connect.js)
- [`test/simple/test-http-after-connect.js`](https://github.com/joyent/node/blob/v0.10.35/test/simple/test-http-after-connect.js)

That way we can test all combinations of (proxy HTTP, HTTPS) x (endpoint HTTP, HTTPS) x (`tunnel` `true`, `false`, `undefined`).

I'm working on it, probably won't be ready until this weekend though.

twolfson:
No worries. This PR isn't blocking me; I have been happily using my fork for the past day.

nylen:
@twolfson version `2.52.0` is out on npm with this change (via #1374).  Let me know if you have any problems using that.

twolfson:
Sweet, thanks!

nylen:
@twolfson use `2.53.0` instead (see #1395, #1396).

Issue #1094, opened 2014-09-24
Request's Architecture


seanstrom:
Hey can some one give me a high level breakdown of request's architecture?
I basically want to get some point of views of how this software is perceived and how it actually works.
@mikeal

Janpot:
Just stumbled across this thread and wanted to share something I worked on a few months ago. I've since abandoned the project, maybe I'll pick it up again some day, maybe not. It's basically a different approach on how to achieve a `request`-like library in a modular way. I started from a base module that implements the most basic stripped down http request I could think of. Then I build up all other functionality in the form of middleware. Just like libraries like express build up the server side functionality of http through middleware functions.
The idea here is that a middleware function receives a `request` object which it can alter at will (headers etc...). Then call a `next` function that calls all downstream middleware up until the basic request. After which the response object bubbles back up where it can be manipulated by the same middleware.
This way I was able to isolate all functionality for e.g. redirection, gzip, cookies into their own modules.
https://github.com/Janpot/kwest
https://github.com/Janpot/kwest-redirect
https://github.com/Janpot/kwest-cookies
https://github.com/Janpot/kwest-gzip
Now, I'm not saying this is a stable library and you should make `request` like this. I just wanted to share a way of looking at it that I didn't yet see in this thread. Which is mimicking the way we build http servers to the way we build http clients.

Glancing over my readmes I notice they are not ready and fully consistent so let me add an example here as well, this is how it worked:

```js
var kwest = require('kwest');
var kwestRedirect = require('kwest-redirect');
var kwestBodyparser = require('kwest-bodyparser');

var request = kwest()
  .use(kwestRedirect({ maxRedirects: 10 }))
  .use(kwestBodyparser());

request.use(function rejectErrors(req, next) {
  // You can also alter req here, like setting headers, etc
  return next(req).then(function (res) {
    // This lib was promise based, but that's ofcourse an implementation detail
    if (res.statusCode !== 200) {
      throw new Error('bad status');
    }
    return res;
  });

});

request({
  url: 'http://www.google.com'
}).then(function (res) {
  // do something with the response here
});
```

Again, not claiming this is better or saying you should do it this way, just want to spark some inspiration by showing a different approach. I found that this approach had its advantages.

I use `request` a lot in my day to day work (and love it) and the reason why I started exploring this `kwest` approach is that I often found myself needing things like "a http request that respects robots.txt" or "a http request that detects encoding and decodes text", etc... which IMO would be great candidates for a middleware/plugin like system.

[...]

PR #1350, opened 2015-01-12
WIP request-oauth as a separate module

simov:
**Not an actual PR**

I got inspired lately from some other refactoring PRs, so I just moved the `oauth` method into a separate _module_. The `oauth` method in request now looks like this:

```js
Request.prototype.oauth = function (_oauth) {
  var self = this

  var result = roauth.oauth({
    uri: self.uri,
    method: self.method,
    headers: self.headers,
    body: self.body,
    qsLib: self.qsLib,
    oauth: _oauth
  })

  if (result.transport === 'header') {
    self.setHeader('Authorization', result.data)
  }
  else if (result.transport === 'query') {
    self.path += result.data
  }
  else if (result.transport === 'body') {
    self.body = result.data
  }

  return self
}
```

My idea is for a separate module in the request org called `request-oauth` currently this is a [single file](https://github.com/simov/request/blob/1bd0ca14b538e0afa3a8066f5f6f5049655461fe/request-oauth.js) with only minor changes in it.

So if @mikeal @nylen @tbuchok @seanstrom and everyone else give it a green light, I would do the following:

1. Create documentation
2. Write unit tests
3. Make some minor changes to the code (right now it's really just copy/pasted)

### DIFF
--- a/request-oauth.js
+++ b/request-oauth.js
@@ -1,0 +1,117 @@
+'use strict'
+
+var querystring = require('querystring')
+var qs = require('qs')
+var caseless = require('caseless')
+var uuid = require('node-uuid')
+var oauth = require('oauth-sign')
+
+/*
+  var args = {
+    uri: {},
+    method: '',
+    headers: {},
+    body: {},
+    qsLib: {},
+    oauth: {}
+  }
+*/
+
+exports.oauth = function (args) {
+  if (!args.uri) {
+    throw new Error('request-oauth: missing url argument')
+  }
+  if (!args.method) {
+    throw new Error('request-oauth: missing method argument')
+  }
+
+  var uri = args.uri
+  var method = args.method
+  var headers = caseless(args.headers || {})
+  var body = args.body || ''
+  var qsLib = args.qsLib || qs
+  var _oauth = args.oauth || {}
+  
+
+  var form, query, contentType = '', formContentType = 'application/x-www-form-urlencoded'
+
+  if (headers.has('content-type') &&
+      headers.get('content-type').slice(0, formContentType.length) === formContentType) {
+    contentType = formContentType
+    form = body
+  }
+  if (uri.query) {
+    query = uri.query
+  }
+
+  var transport = _oauth.transport_method || 'header'
+  if (transport === 'body' && (
+      method !== 'POST' || contentType !== formContentType)) {
+
+    throw new Error('oauth.transport_method of \'body\' requires \'POST\' ' +
+      'and content-type \'' + formContentType + '\'')
+  }
+
+  delete _oauth.transport_method
+
+  var oa = {}
+  for (var i in _oauth) {
+    oa['oauth_' + i] = _oauth[i]
+  }
+  if ('oauth_realm' in oa) {
+    delete oa.oauth_realm
+  }
+  if (!oa.oauth_version) {
+    oa.oauth_version = '1.0'
+  }
+  if (!oa.oauth_timestamp) {
+    oa.oauth_timestamp = Math.floor( Date.now() / 1000 ).toString()
+  }
+  if (!oa.oauth_nonce) {
+    oa.oauth_nonce = uuid().replace(/-/g, '')
+  }
+  if (!oa.oauth_signature_method) {
+    oa.oauth_signature_method = 'HMAC-SHA1'
+  }
+
+  var consumer_secret_or_private_key = oa.oauth_consumer_secret || oa.oauth_private_key
+  delete oa.oauth_consumer_secret
+  delete oa.oauth_private_key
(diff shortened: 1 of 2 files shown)


simov:
Also if that looks just like mindless copy/pasting, I'm still thinking about the re-usability of this piece of code as a separate module, and of course about the exact input/output parameters format. This was just a proof of concept.

seanstrom:
@simov funny thing we had this conversation a while ago in #1094
This would be a good step in cleaning up the code and moving towards a more modular design.
There may be even details about oauth method specifically in that issue.

simov:
@seanstrom I know about this issue, oh wait, it's way too long, but in general I think the `oauth` method can be re-used as a separate module, outside of the request's context.

Only if it have well defined input/output. If you take a look at the _copy/pasted_ code only two of the input arguments are required, the rest have defaults, and I'm really concerned only about the `uri` parameter which requires an `url.parse`'d input. But as I said this was a really quick proof of concept.

seanstrom:
Okay, I see what you mean now.
I think in general we moving pieces of code away from request.js will clean up the code in then long run.
That being said, I am a little wary about throwing errors from the oauth module.
We should implement a try-catch statement or somehow pass the error back up to the user callback, if there's an issue. What do you think?

simov:
Good point, we should probably call `self.callback` or something.

nylen:
Looks good to me.  What is the benefit of having this be a separate module, though, instead of `lib/oauth.js`?

Another thing to consider: in [previous discussion on architecture and refactoring](https://github.com/request/request/issues/1094#issuecomment-58714438) @mikeal suggested a `sign` property that code like this would set to a function:

```js
request.sign = function() {
    // set the signing header / body parameter
};
```

Then, if `self.sign()` is a function, main `Request` would call it right before sending headers.

seanstrom:
I would think the benefits of extracting this out would be it removes more noise from the request.js
Which is important so we can easily and properly refactor that code as well, and we can make this code more maintainable and easily testable. We can even start to test some of the patterns @mikeal mentioned.

What are your guy's thoughts on refactoring this?

simov:
@nylen I think I'm getting the point here. Having a generic lib/auth module with all possible authentications there would be far more powerful I think. I was really looking only at the oauth method.

In that line of thinking probably my approach here is not the best, I'll take a look at the rest of the auth related methods these days and see what I'll come to.

@seanstrom  That's the goal in the long run, exactly!

nylen:
@simov I still think this is a good first step - anything that makes `request.js` shorter is probably a good thing.  I wouldn't combine different auth methods together, instead, have them use a common set of hooks (which don't exist yet, so my comment was just to give some things to think about for the future).

For example, if/when we have a `sign` hook, `request-oauth` could do this:

```js
exports.extend = function(Request) {
  Request.prototype.oauth = function(options) {
    this.sign = function() {
      // add header/querystring/whatever with OAuth signature
    }
  }
}
```

Then, you'd use it like this:

```js
require('request-oauth').extend(Request)
```

simov:
Got it, thanks for the clarification.

If we intend to pass the instance around, I don't think that the removal of any of the `prototype` methods will have a positive impact at this stage, as they are relatively well defined and not so huge.

I think we should start the refactoring on smaller steps in some of the really huge functions and then think about the next move.

The removal of huge amount of code that modifies the state in various ways can even make the refactoring more complicated at this stage.

seanstrom:
We would be passing the constructor around not an instance, i think.
My thoughts are this is low hanging fruit for us to refactor, and we can even start implementing some auth signing patterns. That research alone will help us.

simov:
My idea here was to create a module that has no knowledge of Request, and which is fully encapsulated - receive data and returns data, the modification of the state is carried out in Request. And not the least that can be re-used in another context.

As shown in the code above that somewhat works for the `oauth` method, but then I looked at other authentication methods and could not think of a good way to do the same with them.

For example the `auth` method reads or modifies state on each and every line. The flags set there are then used somewhere else. The approach that I applied here with the `oauth` is not applicable for `auth`, respectively at this stage we can simply copy it outside of request.js and let it modify the state of which we'll not gain anything.

So the removal of the code outside request.js is simple, the question is what we're gaining in terms of maintenance and readability. Currently much easier to check that the `_auth` flag is set in the `auth` method, although there are two other flags set there that are not so obvious, therefore we'll have to search through files. Again not too much of a problem but what's the point?

In the example above I know that it may or may not modify the state, but if you read the mikeal's comment from above that's the design decision that's been made about the state (if not obvious). So we must make sure that there are no too huge functions in request that prevent us from adding and expirimenting with new features. It's not that much about capsulation, because that brings us to much more state and a growing code base. As for various `_vars` that modify the state we should have at least some sort of convention for that before starting to move big chunks of code out of request.

Issue #1122, opened 2014-10-02
Let's spend this money :)


mikeal:
A while back there was a bountysource campaign that brought in over $1,700 for request.

https://www.bountysource.com/teams/request/fundraiser

I haven't been able to figure out the best way to spend this and so far haven't done any bounties. It's probably best to hand this over to all the contributors since I'm a giant bottleneck on this.

nylen:
Hey guys.  Let's spend this money :)

I have access to our Bountysource account now.  I do want us to discuss our process for accepting and paying out changes pretty thoroughly first though.  I am thinking that collaborators should have the option to vote on bounty amounts.  I have not done something like this before, so any suggestions are welcome.

We'll also need to get in touch with our major donors and see what features they'd like, but we want to be wary of letting people buy influence over the direction of the project.

What features are worth paying for?  Some initial ideas:

- $x per % of increased test coverage (to me, this is a no-brainer; it's a very objective measure of our code quality and how well we understand our codebase)
- Browser test suite (hey look @eiriksm :)
- Modularization and refactoring
- Preparing for 3.0 release with streams2

seanstrom:
@nylen I think you outlined some of the big ones that users will really want.
This one may be covered by the 3.0 release or future releases, but what about major API changes that users would want. Perhaps Promise support is something users would like, or just other integrations in general.

seanstrom:
Also Documentation refactoring should be added.
https://github.com/request/request/issues/1340

nylen:
Ok.  What amount do you think is appropriate for test coverage increases?  We're at 88% right now.

I propose that collaborators can vote on issue bounties by saying so in the issue, and we'll take the average if everybody is around the same range.  I'll start with #1366 (I would have done the same for your proxy refactoring too, if we had had this discussion yet).

I'm certainly not opposed to documentation refactoring being added, but I'm not comfortable doing much given that I'm planning on doing the work.  Want to wait until I send the PR (should be sometime this week...) then propose an amount there?

Major new features are good too, but -1 on promise support **specifically** in request (it would probably have to break streams, and there are pretty nice wrapper libraries for it).  We just have to avoid breaking compatibility until we're ready for 3.0.

seanstrom:
Mmm, Well even though its at 88% I would like there to be tests around more of the modules we have been refactoring, and maybe even some clean up and maintenance of the existing test files.
I think that's a good amount of work and could be approximately priced at maybe $100 - $150.
[...]

PR #1366, opened 2015-01-22
Refactor OAuth into separate module

simov:
This is a follow up to my previous PR https://github.com/request/request/pull/1350 Also it might be related to this one https://github.com/request/request/pull/1360

The `getParams` function is about the generation of OAuth parameters. I think it's pretty straightforward to understand which parameters have defaults, which are excluded, and what parameters are required for the OAuth signature.

The `concatParams` function is the old `buildSortedParams` but I though it deserves it's own function, I mean the `oa` parameter was used from the outer scope, so it was not exactly a function in that sense.

One change in `concatParams` is that the signature is no longer appended to the end of the string, but that can be changed if it not confront to the spec in some way.

The `oauth` function is pretty much what's left from the original one, but I think now it's more compact and easy to follow, it even fits on my screen. The `getParams` and `concatParams` are called from there, there is only one level of indirection so again it's very easy to follow.

All of the available input `args` parameters are listed on top of the `oauth` function. Also right now in Request this module is called like `var result = oauth.oauth` which is a bit silly, but I couldn't come up with something more reasonable.

At first I thought that returning of a map is not a good idea for the `oauth` method, but then I liked how it feels, I mean `result.transport` and `result.oauth` sounds good to me.

That's all I could think of at the moment.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -7,7 +7,7 @@ var querystring = require('querystring')
   , oauth = require('oauth-sign')
 
 
-function getParams (_oauth, uri, method, query, form, qsLib) {
+exports.buildParams = function (_oauth, uri, method, query, form, qsLib) {
   var oa = {}
   for (var i in _oauth) {
     oa['oauth_' + i] = _oauth[i]
@@ -49,7 +49,7 @@ function getParams (_oauth, uri, method, query, form, qsLib) {
   return oa
 }
 
-function concatParams (oa, sep, wrap) {
+exports.concatParams = function (oa, sep, wrap) {
   wrap = wrap || ''
   return Object.keys(oa).sort().map(function (i) {
     return i + '=' + wrap + oauth.rfc3986(oa[i]) + wrap
@@ -83,21 +83,21 @@ exports.oauth = function (args) {
       'and content-type \'' + formContentType + '\'')
   }
 
-  var oa = getParams(_oauth, uri, method, query, form, qsLib)
+  var oa = this.buildParams(_oauth, uri, method, query, form, qsLib)
 
   var data
   switch (transport) {
     case 'header':
       var realm = _oauth.realm ? 'realm="' + _oauth.realm + '",' : ''
-      data = 'OAuth ' + realm + concatParams(oa, ',', '"')
+      data = 'OAuth ' + realm + this.concatParams(oa, ',', '"')
       break
 
     case 'query':
-      data = (query ? '&' : '?') + concatParams(oa, '&')
+      data = (query ? '&' : '?') + this.concatParams(oa, '&')
       break
 
     case 'body':
-      data = (form ? form + '&' : '') + concatParams(oa, '&')
+      data = (form ? form + '&' : '') + this.concatParams(oa, '&')
       break
 
     default:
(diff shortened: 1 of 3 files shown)


nylen:
:+1: looks fine to me.  I'm not an expert on OAuth (or our OAuth code) but my test scripts against oauthbin.com still pass so I think moving `oauth_signature` is ok.

I'll merge this in a couple of days if no one objects or beats me to it.

simov:
@nylen I just enabled the `realm` parameter for `query` and `body` transport types. I saw some examples [here](https://groups.google.com/forum/#!topic/twitter-development-talk/KJ47Cio7zUc) so I thought it would be good to make the `realm` parameter available there. Also fixed the test there as it was relying on regular expression.

About the signature position I read [this](http://oauth.net/core/1.0/#anchor14), so although it might work I decided to stick with the spec.

I tested with a dozen of providers using OAuth1, everything seems to be fine.

nylen:
Per discussion in #1122 I'm nominating this for $30 bounty.  @seanstrom @FredKSchott and anyone else, let me know your thoughts on using this a a guinea pig for bounty :)

seanstrom:
Im fine with this being a bounty, good work @simov :+1:

simov:
Not needed at all, but if you insist on giving a bounty for this, I'm fine with your decision. As for the unit testing the new modules, that would come too, still we're going to need our _system_ like tests that involves the entire request object. So for now we can safely rely on what we already have as tests, as I'm not changing or adding behavior during this refactoring, after all it is a refactoring :)

seanstrom:
Oh we will pay you whether you like it or not :).

nylen:
Bountysource says this issue is closed for some reason.  Closing and reopening to see if that fixes it.

simov:
Do I need account on BountySource?

nylen:
yeah, looks like if you log in and go here you may be able to request it:

https://www.bountysource.com/issues/7999758-refactor-oauth-into-separate-module/developers

simov:
ok, posted my claim here https://www.bountysource.com/issues/7999758-refactor-oauth-into-separate-module/developers

nylen:
bleh.  looks like you are all set to be paid $0.  I'll get in touch with bountysource support and see if they can help.

simov:
Probably I have to enter the amount in my comment?

nylen:
Reply from Bountysource support:

---

Hey James,

There's some quirkiness regarding how Bountysource handles GitHub Pull Requests at the moment. All bounties should be on open Issues, not pull requests, so that should be disallowed. There are actually two issues that you've highlighted here:

1) We shouldn't allow bounties on PRs -- We want to encourage people to focus on solving open issues, and not attempt to pay to get specific code push through.
2) We shouldn't allow developers to "claim a bounty" when none exists. There are no backers and there's no money to pay out, so claiming it is pointless.

Is there an issue that corresponds to that open PR (or can you create one)? I would suggest putting the bounty on that issue and, when the issue gets closed the developer can claim the bounty. As a backer of the issue you will have the ability to approve/deny the bounty claim.

Sorry about the rather poor experience in this situation, we have some edge cases to clean up :). Hope this answers your questions.

---

@simov you can claim the bounty on #1379.

nylen:
merging, we'll take care of the bounty at #1379.  Thanks @simov

Issue #1259, opened 2014-11-10
HTTPS over HTTP tunnel: prompted for private key password


brichard19:
Hi,

I'm trying to use HTTPS with client authentication over an HTTP tunnel proxy. On every connection, OpenSSL prompts for the client private key password.

I tried adding:

```
agentOptions {
    passphrase:"password"
}
```

But it does not work when using a proxy.

PR #1368, opened 2015-01-23
[IGNORE] Add missing HTTPS options that were not being passed to tunnel

brichard19:
This change adds some missing HTTPS options that were not being passed to the tunnel module.

This fixes [Issue 1259](https://github.com/request/request/issues/1259) where the user is prompted for the private key password when performing mutual authentication using a tunnel proxy with a password-protected private key.


nylen:
From reading the code, and related work in #1373 I'm not convinced this will actually work as intended:

- https://github.com/request/request/blob/v2.51.0/request.js#L330-L332
- https://github.com/request/request/blob/v2.51.0/request.js#L144-L148
- https://github.com/mikeal/tunnel-agent/blob/v0.4.0/index.js#L119-L126

Can you rebase your PR off #1373 and add some tests?

nylen:
Closing in favor of #1380.

PR #1373, opened 2015-01-24
Rewrite tunnel test to be pure Node.js

nylen:
The new test is doing much more than the old one was, and as far as I can tell, no one was ever running it since it required `squid`.

This will be nice to have when making further changes/fixes related to tunneling and proxies, such as #1344 and #1368.


seanstrom:
:+1:  nice job

seanstrom:
I'm assuming your finished with this, I'll merge tomorrow end of today, unless you say otherwise.

nylen:
thanks :)

PR #1374, opened 2015-01-25
Allow explicitly disabling tunneling for proxied https destinations

### LINKED ISSUE #1293: Disable HTTP tunneling for all protocols

nylen:
Before, tunneling would default to false for http destinations and be
forced to true for https destinations.

Now, allow specifying `tunnel : false` to perform a regular GET request
to HTTPS destinations.  This is insecure, so make a note accordingly.

Supersedes #1344 and fixes #1293.

Also add some more tests for proxy environment variables - it hasn't been easy to test proxying HTTPS URLs without tunneling until this change.

### DIFF
--- a/request.js
+++ b/request.js
@@ -159,7 +159,7 @@ function getTunnelOption(self, options) {
   return undefined
 }
 
-function construcTunnelOptions(request) {
+function constructTunnelOptions(request) {
   var proxy = request.proxy
 
   var tunnelOptions = {
@@ -311,7 +311,7 @@ Request.prototype.setupTunnel = function () {
   proxyHeaderExclusiveList.forEach(self.removeHeader, self)
 
   var tunnelFn = getTunnelFn(self)
-  var tunnelOptions = construcTunnelOptions(self)
+  var tunnelOptions = constructTunnelOptions(self)
 
   self.agent = tunnelFn(tunnelOptions)
   return true
--- a/request.js
+++ b/request.js
@@ -133,6 +133,32 @@ function constructProxyHeaderWhiteList(headers, proxyHeaderWhiteList) {
     }, {})
 }
 
+function getTunnelOption(self, options) {
+  // Tunnel HTTPS by default, or if a previous request in the redirect chain
+  // was tunneled.  Allow the user to override this setting.
+
+  // If self.tunnel is already set (because this is a redirect), use the
+  // existing value.
+  if (typeof self.tunnel !== 'undefined') {
+    return self.tunnel
+  }
+
+  // If options.tunnel is set (the user specified a value), use it.
+  if (typeof options.tunnel !== 'undefined') {
+    return options.tunnel
+  }
+
+  // If the destination is HTTPS, tunnel.
+  if (self.uri.protocol === 'https:') {
+    return true
+  }
+
+  // Otherwise, leave tunnel unset, because if a later request in the redirect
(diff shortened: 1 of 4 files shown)


nylen:
@twolfson can you confirm that this works for your use case?

@simov @seanstrom @FredKSchott review?

twolfson:
Yep, made a gist as a proof of concept. Looks good

https://gist.github.com/twolfson/7b9f4bbe653b47900248

nylen:
one question I still have is how this works when "a previous request in the redirect chain used a tunneling proxy" - I'll add some tests in this PR.

seanstrom:
Ill have time later today to review this, unless someone else beats me to it

nylen:
@seanstrom sounds good, not sure if I'll be able to get to the additional tests today though.

nylen:
@seanstrom this is ready, and I've got another PR based on top of it waiting (#1380 with some cleanups).  think you'll be able to review today?

request.js
```diff
@@ -133,7 +133,7 @@ function constructProxyHeaderWhiteList(headers, proxyHeaderWhiteList) {
     }, {})
 }
 
-function construcTunnelOptions(request) {
+function constructTunnelOptions(request) {
```
seanstrom: good catch

request.js
```diff
@@ -383,8 +377,24 @@ Request.prototype.init = function (options) {
...
+  //    was tunneled
+  //  - Allow the user to override this setting
+  // Implementation:
+  //  - If this is a redirect, self.tunnel may be set -> don't do anything.
+  //  - If options.tunnel is set, use it.
+  //  - Otherwise, tunnel if the destination is HTTPS, but don't set
+  //    tunnel=false if it's not, because if a later rquest in the redirect
+  //    chain is HTTPS then that request (and any subsequent ones) should be
+  //    tunneled.
+  if (typeof self.tunnel === 'undefined') {
```
seanstrom: This nested if statement is the only thing in the code that's catching my eye.
It's not bad by any means, and pretty much necessary given the way we're operating on `self.tunnel`.

However I just wanted to toss this idea in.
What if we used a function that took in self and options, then returned a boolean.
Then in the code we would call that function and assign its value to self.tunnel.

Like this more or less:
```javascript
function setTunnel(self, options) {
  if(self.tunnel) {
    return self.tunnel
  }

  if(options.tunnel) {
    return options.tunnel
  }

  if(self.uri.protocol === 'https:') {
    return true
  }

  return false // not necessary since undefined is falsy
}

// else where in code

self.tunnel = setTunnel(self, options)
```

What do you think?

nylen:
Good idea to make the code cleaner, done.  I tweaked your logic some, luckily I tested the crap out of this behavior so I'm pretty comfortable that it's working as intended.

Let me know if you think this looks OK now, and I'll clean up the history.

seanstrom:
@nylen I think it looks good, thanks for those changes :). And comments thank you for those too.
:+1:

seanstrom:
Ill merge this shortly, unless someone else has some review

nylen:
history cleaned.

PR #1376, opened 2015-01-26
Use karma-browserify for tests. Add browser test coverage reporter.

eiriksm:
Hey.

So I started writing some more browser tests. But before I did that, I wanted to make the test runner to compile browserify by itself, and also add some coverage checking (to make it easier to write tests).

This PR does the following:
- Modifies the scripts in package.json to not compile with browserify directly.
- Modifies the clean script to delete the coverage folder (not sure if this is preferred over adding it to .*ignore)
- Add some new dependencies.
- Make sure karma does both coverage and browserify.
- Also added stderr from the child process that runs karma.

I have some follow up questions about how we should structure the actual test suite, but let's do that in another issue.


nylen:
Can you use `~` instead of `^` in the versions for the new dependencies?

Don't we still need to remove `tests/browser/test-browser.js` in the clean script?

nikku/karma-browserify#91 + a new release of `karma-browserify` should fix the build failure.

eiriksm:
Ah. I didn't even notice the test failed. Sorry. And now that I see it I have actually run into the issue before, hehe. Nice work on patching so quickly!

>Don't we still need to remove tests/browser/test-browser.js in the clean script?
No. test-browser.js will never get built with this change. It was the bundled version of test.js and now if we use karma-browserify it will get built in a temp file in the os temp dir (as you probably got from the referenced issue).

The reason I added rm coverage instead is because there is a js file in there that messes with the linting. Also, we don't want that in the repo or npm. So there's the question: Should we add it to .npmignore .eslintignore and .gitignore?

Updated dependency syntax in a new commit. Will probably fail testing again though, so better leave this open pending the issue above?

For reference: The reason I locked karma-coverage to version 0.2.6 is this issue: karma-runner/karma-coverage#123

nylen:
Yeah, let's ignore it and get rid of the `clean` script.  We shouldn't need to put it in `.eslintignore` because we explicitly specify which paths are to be linted, right?

eiriksm:
Added coverage to .npmignore (it was already in .gitignore). Removed clean script.

You are right, we do specify paths explicitly for eslint, so no worries there.

nylen:
`karma-browserify` 3.0.1 is out, if you can explicitly upgrade to that version then I think we're good to go here.

eiriksm:
done!

nylen:
:+1: thanks!

Issue #1304, opened 2014-12-04
POST 301 redirects don't respect "agent" property


jrasanen:
When POST returns 301, the following GET isn't using the same agent setting.

```js
    var options = {
        url: 'http://website/',
        method: 'POST',
        headers: headers,
        followAllRedirects: true,
        agent: new Socks5ClientHttpAgent({socksHost: '127.0.0.1', socksPort: 9050}),
        form: {
          ...
        }
    }
    // POST returns 301 /index.php
    request.post(options, function (err, httpResponse, body) {...});
```

With this example POST goes through TOR but the following GET doesn't.

polkovnikov-ph:
Have anybody tried fixing that?

mattcg:
I've submitted a pull request that fixes this issue.

mattcg:
@nylen I've tracked down the original issue: https://github.com/request/request/pull/15.

One possible solution is to pass an agent factory, either as an `agentFactory` option or by checking whether `typeof self.agent == 'function'` and if so consider it to be a factory.

nylen:
:+1: nice catch.  I'd take a PR for that (with appropriate tests and docs).  I'd prefer to make it explicit with `agentFactory`.

mattcg:
@nylen there is actually no need for a new option. I realised that the `agentClass` and `agentOptions` options already provide what we need.

The issue is that when `request.init()` is called on [line 141 of redirect.js](https://github.com/request/request/blob/41319044e602af69cdfad9a92dc0c4fd4d30db0a/lib/redirect.js#L141) the options aren't passed and `self.agentClass` is reassigned to the default.

A fix for this that would have no side effects would be to change [line 589](https://github.com/request/request/blob/41319044e602af69cdfad9a92dc0c4fd4d30db0a/request.js#L589) to `if (!self.agent && !self.agentClass) {`.

simov:
Hi @mattcg can you submit a PR for that change, and add some tests for it?

mattcg:
@simov see pull request #1451.

Freyert:
So, this looks to be also connected with issue #1502 or at least stem from a similar issue. It seems like it might be good for `Redirects` to respect the options set in the original request in general. I am probably oversimplifying a bit here though.

In my case the issue is that I've set `request.defaults` with a cookiejar and the `Redirects` do not use said cookie jar.

PR #1378, opened 2015-01-26
Recycle the agent when following a redirect

mattcg:
This fixes https://github.com/request/request/issues/1304. Passes all tests.

### DIFF
--- a/request.js
+++ b/request.js
@@ -1077,7 +1077,6 @@ Request.prototype.onRequestResponse = function (response) {
     // self.method = 'GET' // Force all redirects to use GET || commented out fixes #215
     delete self.src
     delete self.req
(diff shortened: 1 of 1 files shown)


nylen:
Any drawbacks to doing this?  Any thoughts on why it was done this way in the first place?

Can you think of an easy way to add a test for this behavior?

mattcg:
I can't think of any drawbacks.

No idea why it was done in the first place. The message from the commit that added that line offers no explanation.

You mean an easy way to test whether the agent is reused?

nylen:
> You mean an easy way to test whether the agent is reused?

yes.

mattcg:
Closing pending discussion on #1304.

PR #1380, opened 2015-01-26
Add missing HTTPS options that were not being passed to tunnel

brichard19:
Sorry I had to make another pull request - I fudged up when doing a rebase :(

-This change adds some missing HTTPS options that were not being passed to the tunnel module.
-Updated test-tunnel.js to include tests for mutual TLS over HTTP tunnel

This fixes Issue 1259 where the user is prompted for the private key password when performing mutual authentication using a tunnel proxy with a password-protected private key.


nylen:
You have a few lint errors.  Other than that, it looks good at a glance, and I'll do a more detailed review tomorrow.

nylen:
Version `2.52.0` is out on npm with this change (via #1383).

nylen:
Use `2.53.0` instead (see #1395, #1396).

PR #1383, opened 2015-01-28
Add missing HTTPS options that were not being passed to tunnel (@brichard19)

nylen:
This is a re-roll of #1380 with lint issues fixed and a few other cleanups.  It passes more special HTTPS options to the tunnel, but there is more work needed after this:  we still have to do some [special](https://github.com/nylen/request/blob/0f6d1b1bfde40e8baf463d6ff8227b43876af6d7/tests/test-tunnel.js#L32-L36) [stuff](https://github.com/nylen/request/blob/0f6d1b1bfde40e8baf463d6ff8227b43876af6d7/tests/test-tunnel.js#L114-L124) to get HTTPS tunnels and destinations to use a custom certificate authority.  We will probably need to modify [`tunnel-agent`](https://github.com/mikeal/tunnel-agent) in the future to fix this.

I'd like to release a new version after this change gets through, it has been a while and we have a couple of fixes that people have requested.


nylen:
@seanstrom @simov @FredKSchott review pls?

seanstrom:
This seems okay, I believe this has reviewed throughly else-where correct?

nylen:
not sure about that, it's a pretty simple change though, and it includes a decent test case.

seanstrom:
ill give another look over today, then ill merge

nylen:
merging and releasing new version

PR #1387, opened 2015-01-29
Clean up debug code

nylen:
Make `debug` a prototype method.  This has a couple of advantages:

- It's no longer necessary for the debug code to be at a specific place
  in `request.js`, see #1354
- It's easier for helper methods in other files to generate debug
  messages, see #1360

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -8,8 +8,9 @@ var md5 = helpers.md5
   , toBase64 = helpers.toBase64
 
 
-function Auth () {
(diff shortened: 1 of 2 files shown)


simov:
@nylen with this changes https://github.com/request/request/pull/1407 we'll be able to add the debug code as you proposed. We can then access it like

```js
request.debug()
```

inside each method. All of the debug calls are currently commented out in https://github.com/request/request/pull/1407 so you can easily find them.

PR #1388, opened 2015-01-29
Upgrade mime-types package version

roderickhsiao:
the dependency 
``form-data`` is using ``"mime-types": "~2.0.3"``
https://github.com/felixge/node-form-data/blob/master/package.json#L20

but in ``request`` we are using ``~1.0.1`` which causes multiple version of copy being download whenever we require ``request``


nylen:
:+1: thanks!

PR #1392, opened 2015-02-02
Improve `timeout` option description

watson:
Emphasise when the OS-wide TCP timeout overrules the `timeout` option. The previous explanation could leave the reader thinking that the OS-wide timeout would overrule a bigger `timeout` option even if the TCP connection was successfully established


nylen:
:+1: thanks!

Issue #1395, opened 2015-02-02
JSON POST body URI encoded (RFC3986)


chacal:
Due to pull request [#1315](https://github.com/request/request/pull/1315) all POSTed JSON content is  URI encoded in 2.52.1.

Is there some way to avoid this? The server I'm POSTing to does not expect JSON data to be encoded, but instead sent as is.

Anyhow, is it OK to encode JSON content at all as the format itself allows the usage of e.g. '!' and '*' without any encoding? `{"key": "value !"}` is a very different JSON than `{"key": "value %21"}`

nylen:
Thanks for reporting, I agree that we should not pass JSON POST bodies through `rfc3986`.  @simov want to take a look so we can release a fixed version quickly?

simov:
That was the reason I proposed this change as an option (first commit). If Twitter allows json post bodies (not sure) then there it should be rfc3986 encoded. So what's the expected behavior exactly?

nylen:
Encoding regular `application/x-www-form-urlencoded` POST bodies is transparent: if we encode too much, the server will decode it.

I didn't think about this until now, but encoding JSON in this way is **not** transparent.  I'd be surprised if there were any servers that will decode JSON in this way.

Propose deleting [this line](https://github.com/request/request/blob/v2.52.0/request.js#L1432), let me know what you think.

simov:
OK, I'll take a look, and submit PR (the tests should be changed a bit)

chacal:
Shouldn't the encoding be removed also on line 1439? There we are setting JSON request body as well, I believe.

simov:
I'm just looking at it :)

nylen:
Version `2.53.0` is on npm with this fix.

chacal:
Thanks for the very fast response on this! Version 2.53.0 works like a charm! :)

nylen:
glad to hear it, that was a nasty bug :sweat:

crobinson42:
Hey @chacal @nylen I'm running into this using `request` to post to github using the github api. It's encoding my JSON and I can't seem to understand a way around it. Any help?

I send this `{ "title": "Test Bug Title..Kaboom"}` and github receives this `{ "title": "Test Bug Title..Kaboom%21" }`

```
request
				.post({
					url: "https://api.github.com/repos/crobinson42/therms-public/issues",
					form: JSON.stringify(githubIssueObj),
					headers: {
						'User-Agent': 'App Auto Bug Filing',
[...]

PR #1396, opened 2015-02-02
Do not rfc3986 escape JSON bodies

simov:
Example `{json:{}}` or `{body:{}, json:true}`
Related to https://github.com/request/request/pull/1315
Closes https://github.com/request/request/issues/1395

Also take a look at the [options section](https://github.com/request/request#requestoptions-callback) of the documentation.

### DIFF
--- a/request.js
+++ b/request.js
@@ -1428,15 +1428,15 @@ Request.prototype.json = function (val) {
     if (self.body !== undefined) {
       if (!/^application\/x-www-form-urlencoded\b/.test(self.getHeader('content-type'))) {
         self.body = safeStringify(self.body)
+      } else {
+        self.body = rfc3986(self.body)
       }
-      self.body = rfc3986(self.body)
       if (!self.hasHeader('content-type')) {
         self.setHeader('content-type', 'application/json')
       }
     }
   } else {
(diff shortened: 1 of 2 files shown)


nylen:
The behavior looks correct.  I tried to clean up the test a little to get comfortable that we're doing the right thing here, can you pull in https://github.com/nylen/request/commit/29f98f31ed1c90a08786e7c8345eaf0bbd289877?

request.js
```diff
@@ -1428,15 +1428,15 @@ Request.prototype.json = function (val) {
     if (self.body !== undefined) {
       if (!/^application\/x-www-form-urlencoded\b/.test(self.getHeader('content-type'))) {
         self.body = safeStringify(self.body)
+      } else {
+        self.body = rfc3986(self.body)
```
nylen: Can you swap this conditional around?  Also, I see we have a test for it (`body + header + json`) but in what situation  is this needed?

simov:
Done. Are we good with this one (regarding your comment from above)?

nylen:
Yeah, thanks for taking a look!

nylen:
version `2.53.0` released.

PR #1406, opened 2015-02-05
Refactor redirect logic into separate module

simov:
In the last few days I have the idea that the request plugins should introduce `onRequest` and `onResponse` hooks at minimum. Also all of them should be classes that caches the `request` instance internally.

Then in `base` we should have defaults like:

```js
function Auth () {
  
}
Auth.prototype.onRequest = function (options) {
  // just nothing, or throw some error accordingly
}
Auth.prototype.onResponse = function (response) {
  // just nothing, or throw some error accordingly
}

function Redirect () {
  
}
Redirect.prototype.onRequest = function (options) {
  // just nothing, or throw some error accordingly
}
Redirect.prototype.onResponse = function (response) {
  // just nothing, or throw some error accordingly
}
```
and so on

The idea is that request `base` should work without having certain plugins included. Essentially all plugins should modify the request instance on their own.

Then we can configure the plugins like this:

```js

// just like the defaults method
var custom = request.config(['auth', 'redirect', 'streams'])
```

Obviously with this approach we should have the hooks defined internally, in other words this does not allow for user made plugins. The reason why is that certain plugin hooks should be called in specific parts of the code, like for example the `auth` response hook called inside the `redirect` plugin (but that may change)

My initial idea was to just loop through all plugins required by the config method, inside the `init` and `onRequestResponse` methods, so we can think about these plugins as middlewares that are used or not. I guess that would be possible at some point, and this will allow just about any custom made plugin/middleware to be called for each request/response pair, and modify the state accordingly.

Let me know what do you think :)

### DIFF
--- a/lib/redirect.js
+++ b/lib/redirect.js
@@ -1,0 +1,146 @@
+'use strict'
+
+var url = require('url')
+var isUrl = /^https?:/
+
+function Redirect (request) {
+  this.request = request
+  this.followRedirect = true
+  this.followRedirects = true
+  this.followAllRedirects = false
+  this.allowRedirect = function () {return true}
+  this.maxRedirects = 10
+  this.redirects = []
+  this.redirectsFollowed = 0
+}
+
+Redirect.prototype.onRequest = function () {
+  var self = this
+    , request = self.request
+
+  if (request.maxRedirects !== undefined) {
+    self.maxRedirects = request.maxRedirects
+  }
+  if (typeof request.followRedirect === 'function') {
+    self.allowRedirect = request.followRedirect
+  }
+  if (request.followRedirect !== undefined) {
+    self.followRedirects = !!request.followRedirect
+  }
+  if (request.followAllRedirects !== undefined) {
+    self.followAllRedirects = request.followAllRedirects
+  }
+  if (self.followRedirects || self.followAllRedirects) {
+    self.redirects = self.redirects || []
+  }
+}
+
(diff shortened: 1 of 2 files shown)


lib/redirect.js
```diff
@@ -0,0 +1,146 @@
...
+
+var url = require('url')
+var isUrl = /^https?:/
+
+function Redirect (request) {
+  this.request = request
+  this.followRedirect = true
+  this.followRedirects = true
+  this.followAllRedirects = false
+  this.allowRedirect = function () {return true}
```
seanstrom: We probably shouldn't be defining this for every instance, given it just defaults to true.
You should probably define this has a named function and then assign this.allowRedirect to the function.

seanstrom:
This code seems good. Like you mentioned, this is mostly refactoring the code to other modules.
Which is awesome by the way :+1: 
Overall I would say I don't have much feedback on how the plugin system is all going to work, but that these changes probably get us closer to our goals then farther away.
Again good job!

lib/redirect.js
simov: I'm adding all of the object fields in the ctor, just like the modules are required on top, so that you know what's used and what's available. It's like a convention or something. Plus this module will definitely get refactored a bit more, at some point.

simov:
@seanstrom exactly my point :+1: The name of the other PR is a bit misleading (Plugin System), I just wanted to point out that I'm refactoring the rest of the modules so that they confront to what I said from above.

PR #1407, opened 2015-02-05
Plugin system

simov:
Closes #1400 
Closes #1406 

I just merged https://github.com/request/request/pull/1400 and https://github.com/request/request/pull/1406 here + refactored all of the 4 plugins to use similar structure.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -8,8 +8,9 @@ var md5 = helpers.md5
   , toBase64 = helpers.toBase64
 
 
-function Auth () {
+function Auth (request) {
   // define all public properties here
+  this.request = request
   this.hasAuth = false
   this.sentAuth = false
   this.bearerToken = null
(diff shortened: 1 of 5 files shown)


nylen:
I like the concept.  I'm a little worried about unexpectedly breaking things though, like #1407.

I have been swamped lately (just accepted a job with Automattic!) so I'll need at least a few days to do a proper review.  @mikeal @FredKSchott @seanstrom etc. any help appreciated.

simov:
I'm really not changing that much at this stage, just moving code around, the instance bug with auth is very silly, but apparently not caught by any of the tests.

nylen:
I agree, I feel like I could have caught that in reviewing though, so I'd like to be more thorough this time.

simov:
OK, my point in https://github.com/request/request/issues/1409 was that it's about code modified in this PR + I already merged two other PR's here, so I don't want to push even more things here :)

mikeal:
I don't have time to do a full review but I will say that I like this approach. I'll be impressed if you can actually work the forwarding logic in to a plugin, that logic always seemed to infect parts of the code I never thought I'd have to worry about :)

simov:
@mikeal I'm not familiar with the entire code base in details tbh, but still the refactoring should start from somewhere.

lalitkapoor:
@nylen off topic, but congrats on the new job!! :)

seanstrom:
I'll be doing a thorough review today :)

simov:
Thanks @seanstrom take a look at this PR https://github.com/request/request/pull/1406 too, as most of my latest ideas are there. Also not really that much changed at this point, I'm really just moving code into separate scopes as I see fit. For example the last one - the redirect one is just something that I thought it deserves it's own scope, and in fact except two of the fields in this newly created class are used externally, so it looks like a separate entity after all.

Issue #411, opened 2013-01-10
Header key 'host' should be 'Host'


durango:
Some servers, such as FedEx's API servers, actually care about this case sensitivity.

obender:
Hi, 
We have same issue vs the content-length, Apache Tomcat HTTPD lib refses to accept the low capital header... content-length its' allowing  in Content-length only...

mikeal:
The standard in Node.js is lowercase. In fact, core http automatically lowers headers in the server.

Apache Tomcat HTTPD is violating the HTTP spec which explicitly states that headers are caseless. We do a lot of work in `request` to ensure we are caseless, I even wrote this entire dependency just to deal with it https://www.npmjs.com/package/caseless :)

noranana:
Apache Tomcat HTTPD is violating the HTTP spec????

PR #1408, opened 2015-02-05
'Authorization' header needs to have initcap

AmedeeBulle:
Some servers won't authenticate with 'authorization' header, it needs to be 'Authorization'...

### DIFF
--- a/request.js
+++ b/request.js
@@ -1059,7 +1059,7 @@ Request.prototype.onRequestResponse = function (response) {
   } else if (response.statusCode === 401) {
     var authHeader = self._auth.response(self.method, self.uri.path, response.headers)
     if (authHeader) {
-      self.setHeader('authorization', authHeader)
+      self.setHeader('Authorization', authHeader)
       redirectTo = self.uri
     }
   }


nylen:
-1 on forcing the header to init-caps, servers that require this are in clear violation of the spec.  Also the Node.js convention for HTTP headers is lower-case.

We've had similar issues with the `Host` header (see #1211).  The approach we took there is to allow users to specify a `Host` header manually, and if they do, remember the case they used across redirects.  I like this better because it calls attention to the broken servers and encourages them to be fixed.

Try specifying `{ headers : { 'Authorization' : 'dummy' } }` in your request options along with your auth credentials.  The module we use to manage HTTP headers, [`caseless`](https://github.com/request/caseless), will preserve header case if a header already exists.  This should work for now (perhaps not across redirects).

+1 to documenting and testing a similar approach going forward, can you create a separate issue on this for further discussion?

AmedeeBulle:
@nylen tnx.

Note that this trick does __not__ work for redirects as the auth header is cleared after the first request.

Also there are a couple of other places in request.js where Authorization is with init-cap ;)

nylen:
Interesting find.  I try to follow precedent set before I was involved with this project, for example https://github.com/request/request/issues/411#issuecomment-53656998.

PR #1413, opened 2015-02-07
Fix basic auth

### LINKED ISSUE #1409: Redirect is loosing Authorization
After a page redirect, authorization data is lost.

This is because we have an unconditional 'new Auth' in init() at this line: https://github.com/request/request/blob/master/request.js#L498

The following hack works for 'Basic', but we probably need something more elaborate for the other authorization schemes.
```diff
--- request.js.orig	2015-02-05 14:58:12.401690494 +0100
+++ request.js	2015-02-05 14:57:24.809909090 +0100
@@ -495,6 +495,7 @@
   }
 
   // Auth must happen last in case signing is dependent on other headers
+if(!self._auth ) {
   self._auth = new Auth()
 
   if (options.oauth) {
@@ -528,7 +529,7 @@
       options.auth.bearer
     )
   }
-
+}
   if (self.gzip && !self.hasHeader('accept-encoding')) {
     self.setHeader('accept-encoding', 'gzip')
   }
```
(This feature is needed for single sign on, when authorization is done by a third party through redirects)

### LINKED ISSUE #1412: Digest authentication fails if http method is provided via options
I'm building a client/server application with nodejs. The server uses express and passport and has API protected with Diget authentication. The client uses request.

The following operation is unable to authenticate:

```javascript
  request({
    method: 'get',
    uri: 'http://127.0.0.1:8066/api/test',
    auth: {
      user: 'myuser',
      pass: 'mypass',
      sendImmediately: false
    }
  }, function(err, resp, body) {
    console.log(resp.statusCode);
    console.log(body);
  });
```

However, the following code works:

```javascript
  request.get({
    uri: 'http://127.0.0.1:8066/api/test',
    auth: {
      user: 'myuser',
      pass: 'mypass',
      sendImmediately: false
    }
  }, function(err, resp, body) {
    console.log(resp.statusCode);
    console.log(body);
  });
```

I was expecting the operations to be equivalent. Am I doing something wrong?

simov:
Closes #1400 
Closes #1406 
Closes #1407 

Fixes #1409
Fixes #1412 

Again on top of the previous 3 PRs. @nylen I think we should release a new version after this fix goes in.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -8,8 +8,9 @@ var md5 = helpers.md5
   , toBase64 = helpers.toBase64
 
 
-function Auth () {
+function Auth (request) {
   // define all public properties here
+  this.request = request
   this.hasAuth = false
   this.sentAuth = false
   this.bearerToken = null
@@ -108,11 +109,28 @@ Auth.prototype.digest = function (method, path, authHeader) {
   return authHeader
 }
 
-Auth.prototype.response = function (method, path, headers) {
+Auth.prototype.onRequest = function (user, pass, sendImmediately, bearer) {
   var self = this
+    , request = self.request
+
+  var authHeader
+  if (bearer !== undefined) {
+    authHeader = self.bearer(bearer, sendImmediately)
+  } else {
+    authHeader = self.basic(user, pass, sendImmediately)
+  }
+  if (authHeader) {
+    request.setHeader('authorization', authHeader)
+  }
+}
+
+Auth.prototype.onResponse = function (response) {
+  var self = this
+    , request = self.request
+
   if (!self.hasAuth || self.sentAuth) { return null }
 
-  var c = caseless(headers)
+  var c = caseless(response.headers)
 
   var authHeader = c.get('www-authenticate')
   var authVerb = authHeader && authHeader.split(' ')[0].toLowerCase()
@@ -126,7 +144,7 @@ Auth.prototype.response = function (method, path, headers) {
       return self.bearer(self.bearerToken, true)
 
     case 'digest':
-      return self.digest(method, path, authHeader)
+      return self.digest(request.method, request.path, authHeader)
   }
 }
 
--- a/lib/multipart.js
+++ b/lib/multipart.js
@@ -1,0 +1,104 @@
+'use strict'
+
+var uuid = require('node-uuid')
+  , CombinedStream = require('combined-stream')
+  , isstream = require('isstream')
+
+
+function Multipart () {
+  this.boundary = uuid()
+  this.chunked = false
+  this.body = null
+}
+
+Multipart.prototype.isChunked = function (request, options) {
+  var chunked = false
+    , parts = options.data || options
+
+  if (!parts.forEach) {
+    throw new Error('Argument error, options.multipart.')
+  }
+
+  if (request.getHeader('transfer-encoding') === 'chunked') {
+    chunked = true
+  }
+
+  if (options.chunked !== undefined) {
+    chunked = options.chunked
+  }
+
+  if (!chunked) {
+    parts.forEach(function (part) {
+      if(typeof part.body === 'undefined') {
+        throw new Error('Body attribute missing in multipart.')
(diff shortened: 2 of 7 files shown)


simov:
@nylen I think only the last commit here needs review, since @seanstrom reviewed the rest of the commits.

FredKSchott:
@simov this looks awesome, sorry I couldn't help review sooner (like @nylen I'm in the middle of switching jobs). Love the move to modularize but I'm still super wary to give every plugin a blank check to monkey patch the request object. Understanding the request state and chasing bugs around is going to be much more difficult for everyone after a change like this (and it was already pretty difficult before). 

Sounds like I'm late to the review party, so don't let me get in the way of merging this. But I'd still like us to look into how we can build some structure/rules into this plugin system to help protect contributors from introducing conflicts all willy-nilly :)

simov:
Think about these 4 request modules as Express middlewares, where `res.locals` is called `self.request` in our case. Plus that may change, but we can't just magically switch from code base written in one file to something perfect, unless we start a new project from scratch probably. The refactoring should be made over the time, and since it was not, we'll need some time to get there.

simov:
Anyone I think we should merge this soon, @nylen ?

FredKSchott:
The express middleware analogy might be where I'm having trouble with this then. The idea of functions sending request/response state down a middleware waterfall is very different from plugins all modifying the same parent object. While the potential for conflicts are present in both, middleware is very linear (one direction) and restricted (req/res state only) while this new system goes around in different loops & paths (redirects, gzip, etc.) and has no restrictions (each can modify whatever on the request object).

I'm not advocating for more work, just that we do that work in a thoughtful way. This is a really good first step, and I'd much rather see this than nothing. But I've been burned by this thinking almost every time, where some half-measure is implemented as a first step but the second, third, etc. steps never actually happen. I'm only asking that we spend some time now or immediately after merging to think about how this system could be improved to be safer and more maintainable.

simov:
I know about the state problem, in fact everyone does, and I was the one advocating for proper capsulation in the first two modules https://github.com/request/request/pull/1360 and https://github.com/request/request/pull/1366 Also I'm not saying that I won't continue with the refactoring, I'm just presenting it at steps that are more easy to review. So the idea is pretty simple - I have a huge amount of code in one file (around 2000 lines at start), then the first step is to move code out of it, before I start to see a pattern. Btw if you take a look at these 4 new files (probably except the redirect one) they actually look pretty logical, and are very easy to read, so guess what, even this simple step took a few hours to make. My point here is that I think about this stuff while doing it, and that's completely different than only thinking about it.

Unfortunately design patterns and principles are not applicable in every situation. This project strives to have the shorter possible code base, without any abstraction. Plus it read/write the state so often, that it makes the proper capsulation worthless - essentially you end up with a bunch of copy state on the one and on the other side of the module, which certainly doesn't look good, nor making this project easier to maintain.

The final goal of this refactoring is to make the request instance configurable, but before we start to think about that, we actually need that code out in modules, and right now I'm seeing this as middlewares that modifies the state of the request object, because essentially this is what happens. So again - that's not the final look of the code base, after all we hope it will live longer :)

FredKSchott:
@simov I didn't mean to offend you, I'm sorry if I did. I'm not trying to minimize the size of the problem you're tackling here, or the amount of effort it requires. I just want to make sure we all understand and have thought through the concerns that come with this chosen path. It sounds like you have, so my concerns are alleviated.

nylen:
@FredKSchott we have a common naming convention of `onRequest` and `onResponse` now, so IMO the next step is to register everyone that wants to use these hooks, then call them all in the same place.  I think that helps address the state concerns, thoughts?

@simov thanks again for taking this on, I'm sorry it has been sitting for so long.  In your opinion where do we go from here?

nylen:
oh and @FredKSchott congrats! where are you going to?

simov:
@FredKSchott I'm not offended at all :) I'm sure we'll continue this discussion in the next PRs to come

PR #1415, opened 2015-02-08
Add node.js 0.12 and io.js to travis.yml

demohi:


simov:
Apparently it won't be as straightforward as this.

FredKSchott:
A lot of request logic was changed during the last two years in v0.11, I'm not suprised :/

nicolasmccurdy:
Node 0.12 and io.js support seems like it's both important to this project and at high demand. I think it would benefit this project to merge this pull request into a branch that would then be used for 0.12 and io.js development (so everyone can test against 0.12 and io.js in CI without breaking master's build).

nicolasmccurdy:
By the way, does anyone know if it's possible to allow failures for specific Node versions on Travis? It has something like that for environment variables, but if we can figure out a way to get this to work for Node versions we can show 0.12 and io.js results in CI but prevent the build from being marked as failing while the fixes are in development

nicolasmccurdy:
To answer my own question, yes, it is possible to allow failures on specific versions of Node. I have created a pull request similar to yours, but with allowed failures: #1453.

Issue #1416, opened 2015-02-08
Content-Type header incomplete for "multipart/related" requests


pronvit:
From RFC:

```
3.1.  The Type Parameter

   The type parameter must be specified and its value is the MIME media
   type of the "root" body part.  It permits a MIME user agent to
   determine the content-type without reference to the enclosed body
   part.  If the value of the type parameter and the root body part's
   content-type differ then the User Agent's behavior is undefined.

3.2.  The Start Parameter

   The start parameter, if given, is the content-ID of the compound
   object's "root".  If not present the "root" is the first body part in
   the Multipart/Related entity.  The "root" is the element the
   applications processes first.
```

simov:
That citation is nice, but can you give us an example, also you can set your multipart header by yourself (if that works for you)

```js
request({headers:{'content-type':'multipart/related; something=else'}})
```

pronvit:
In my case it has to be

```
Content-Type: multipart/related; boundary=XXX; type=text/xml; start="<root>"
```

Where `text/xml` and `<root>` are content-type and content-id of the main part (it's a soap request with attachment).

According to https://github.com/request/request/blob/master/request.js#L1386 if I pass content-type myself, it will take only substring before the first ';', which is effectively `multipart/related` only, so it won't help. I don't understand the purpose of that line completely, actually.

simov:
Got it, it's a bug. I will add a failing test case and a fix once https://github.com/request/request/pull/1413 got in.

pronvit:
Thanks.

nylen:
that link should be https://github.com/request/request/blob/v2.53.0/request.js#L1386 (links to specific lines in master tend to break quickly)

PR #1423, opened 2015-02-12
Allow fully qualified multipart content-type header

simov:
Fixes https://github.com/request/request/issues/1416

The problem here was with poorly written logic about the boundary field in multipart/related content-type header. Now we allow fully qualified multipart headers, including ones containing their own boundary, in that case the boundary property is set from the content-header for that instance (see test-multipart.js)

I've done some major refactoring on the chunked detection tests and moved them to test-multipart-chunked.js I also fixed the logic where the transfer-encoding header overrides the value set through the chunked key, because the header wasn't removed anyway, and I don't think that should be the case either. Probably the previous tests were not catching this edge case.

### DIFF
--- a/lib/multipart.js
+++ b/lib/multipart.js
@@ -21,12 +21,12 @@ Multipart.prototype.isChunked = function (options) {
     throw new Error('Argument error, options.multipart.')
   }
 
-  if (self.request.getHeader('transfer-encoding') === 'chunked') {
-    chunked = true
-  }
-
   if (options.chunked !== undefined) {
     chunked = options.chunked
+  }
+
+  if (self.request.getHeader('transfer-encoding') === 'chunked') {
+    chunked = true
   }
 
   if (!chunked) {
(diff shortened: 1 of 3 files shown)


simov:
@pronvit can you test out this branch and see if it works for your case?

lib/multipart.js
```diff
@@ -51,11 +51,16 @@ Multipart.prototype.setHeaders = function (chunked) {
...
-  var contentType = (!header || header.indexOf('multipart') === -1)
-    ? 'multipart/related'
-    : header.split(';')[0]
 
-  self.request.setHeader('content-type', contentType + '; boundary=' + self.boundary)
+  if (!header || header.indexOf('multipart') === -1) {
+    self.request.setHeader('content-type', 'multipart/related; boundary=' + self.boundary)
+  } else {
+    if (header.indexOf('boundary') !== -1) {
+      self.boundary = header.replace(/.*boundary=([^\s;])+.*/, '$1')
```
assaf: `([^\s;])+` picks the first character of the boundary, should be `([^\s;]+)`

simov: Thanks for pointing that out @assaf the bug was fixed here https://github.com/request/request/pull/1505 I had a faulty test for that case.

PR #1426, opened 2015-02-12
Fixing tests to pass on io.js

mikeal:
This fixes all the tests but test-https, which is so convoluted I'm having a hard time figuring out what is broken.


FredKSchott:
:+1:  but tests are now failing on node as well

simov:
I think we should merge at least the passing test fixes into master

cc @nicolasmccurdy the only failing test on node 0.12/iojs is the strict branch in test-https.js

simov:
It's not fixed, I just merged master so that the work can continue here (the certificates were expired) Once the failing test is fixed we should fix the readme and the travis.yml

nicolasmccurdy:
Can we add Node 0.12 and io.js to the Travis config (see #1415) so we can see the results of these fixes in CI ?

nicolasmccurdy:
> I think we should merge at least the passing test fixes into master

@simov Agreed.

simov:
Ok, I'm going to merge the passing test from this branch, then it's probably a good idea to have a separate issue only for the failing https test, and put a bounty on it @nylen @mikeal ?

simov:
Merged. Check your PRs from small merge conflicts (if any)

PR #1428, opened 2015-02-13
Deprecate Node v0.8.x

### LINKED ISSUE #1427: Deprecate 0.8 support
The policy so far has been to only support the last two stable versions of Node.

Deprecating 0.8 will make it easier for contributors to get the tests passing and make it much easier to move to new style streams.

nylen:
Closes #1427

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
-  - "0.8"
   - "0.10"
-before_install: npm install -g npm@~1.4.6
 after_script: ./node_modules/.bin/istanbul cover ./node_modules/tape/bin/tape tests/test-*.js --report lcovonly && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js --verbose
 webhooks:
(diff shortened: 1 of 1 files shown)


FredKSchott:
:+1: This would also mean we're no longer stuck with a super old version of eslint

tikotzky:
:+1: but you should probably update the `engines` property in `package.json` to note that its not supported.

something like: ` "engines" : { "node" : ">=0.10 <0.12" }`

rlidwka:
> something like: `"engines" : { "node" : ">=0.10 <0.12" }`

You'll cut off io.js this way.

tikotzky:
@rlidwka right now iojs is not supported anyway, and  `engines` is just advisory so it wouldn't actually block you from installing and using on iojs.

nylen:
@rlidwka ... can you propose a superior alternative?

rlidwka:
> can you propose a superior alternative?

Maybe just use `{ "node": ">= 0.10" }`?

nylen:
The problem with that is not everything works with `0.12` yet.  Let's discuss any changes to `engines` in a separate issue.

PR #1429, opened 2015-02-15
Throw error when making HEAD request with a body

tikotzky:
Currently it only throws the error if you make the request using `request.head(uri)`

This PR updates to also throw the error when making the request by manually setting the method

```javascript
request({
  uri: uri,
  method: 'HEAD'
})
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,6 +47,10 @@ function request (uri, options, callback) {
   options.callback = params.callback
   options.uri = params.uri
 
+  if (params.options.method === 'HEAD' && paramsHaveRequestBody(params)) {
+    throw new Error('HTTP HEAD requests MUST NOT include a request body.')
+  }
+
   return new request.Request(options)
 }
 
@@ -66,11 +70,6 @@ request.get = function (uri, options, callback) {
 request.head = function (uri, options, callback) {
   var params = initParams(uri, options, callback)
   params.options.method = 'HEAD'
-
-  if (paramsHaveRequestBody(params)) {
-    throw new Error('HTTP HEAD requests MUST NOT include a request body.')
-  }
-
   return requester(params)(params.uri || null, params.options, params.callback)
 }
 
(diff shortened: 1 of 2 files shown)


tikotzky:
This change is required in order to fix recursive `request.defaults` requesters.

I'm waiting until this gets merged before opening that PR since that change needs the changes from this PR.

If you would like to see the changes to fix recursive `request.defaults` requesters they are here:
https://github.com/tikotzky/request/commit/e2f23386a415a60016f2b09989d89975acdca256
https://github.com/tikotzky/request/commit/6330b9a81810941724a407b15e87e66b92c58af4

nylen:
Looks good to me.  Thanks!

gabmontes:
@tikotzky if `request` is called with a callback, it should call the callback with an error instead of throwing to keep the interface consistent. Currently you have to handle errors this way:

```javascript
try {
  request({
    uri: uri,
    method: 'HEAD'
  }, function (err, callback) {
    // manage standard success and error cases
  })
} catch (err) {
  // manage throw case
}
```

Wouldn't it be much cleaner to only manage the errors within the callback?

tikotzky:
@gabmontes I don't remember exactly whats going on here.

But i'm pretty sure the reasoning is that if you pass a body to a `HEAD` request then its an invalid request that cannot be made. Since the lib cannot make the request it throws.

The callback will get called with an error in the case that there was an error while making the request. If you try to make an invalid request the lib will just throw.

gabmontes:
Thanks for your quick response, @tikotzky!!!

I just opened #2109 as this should be changed to always call the callback instead of throwing errors. Please consider doing that change as it will clean up the error handling code in user-space.

In the meantime, a quick workaround I implemented is something like:

```js

function safeRequest (options, callback) {
  try {
    request(options, callback);
  } catch (err) {
    callback(err);
  }
}
```

Thanks again!

PR #1430, opened 2015-02-16
Fix recursive requester

tikotzky:
Right now if you set a `requester` function in `request.defaults` it will override all further defined `requester` functions when using `request.METHOD` to make a request. For example

```javascript
var request1 = request.defaults({}, function(uri, options, callback){
	console.log('I AM THE FIRST REQUESTER FUNCTION');
	request(uri, options, callback);
})

var request2 = request1.defaults({}, function(uri, options, callback){
	console.log('I AM THE SECOND REQUESTER FUNCTION');
	console.log('I WILL NEVER GET CALLED BY request2.METHOD');
	request1(uri, options, callback);
})

// currently using request.METHOD will only hit the first requester
request2.get(url)
// I AM THE FIRST REQUESTER FUNCTION

// using request directly will hit the both requesters
request({
  uri: url,
  method: 'GET'
})
// I AM THE SECOND REQUESTER FUNCTION
// I WILL NEVER GET CALLED BY request2.METHOD
// I AM THE FIRST REQUESTER FUNCTION
```

This PR fixes `request.defaults` so that recursive `requester` functions will get called by both `request()` and `request.METHOD()`

### DIFF
--- a/tests/test-defaults.js
+++ b/tests/test-defaults.js
@@ -213,6 +213,35 @@ tape('recursive defaults', function(t) {
   })
 })
 
+tape('recursive defaults requester', function(t) {
+  t.plan(4)
+
+  var defaultsOne = request.defaults({}, function(options, callback) {
+      var headers = options.headers || {}
+      headers.foo = 'bar1'
+      options.headers = headers
+
+      request(options, callback)
+    })
+    , defaultsTwo = defaultsOne.defaults({}, function(options, callback) {
+      var headers = options.headers || {}
+      headers.baz = 'bar2'
+      options.headers = headers
+
+      defaultsOne(options, callback)
+    })
+
+  defaultsOne.get(s.url + '/get_recursive1', function (e, r, b) {
+    t.equal(e, null)
+    t.equal(b, 'TESTING!')
+  })
+
+  defaultsTwo.get(s.url + '/get_recursive2', function (e, r, b) {
+    t.equal(e, null)
+    t.equal(b, 'TESTING!')
+  })
+})
+
 tape('test custom request handler function', function(t) {
   t.plan(2)
 
--- a/index.js
+++ b/index.js
@@ -54,47 +54,40 @@ function request (uri, options, callback) {
   return new request.Request(options)
 }
 
-function requester(params) {
-  if(typeof params.options._requester === 'function') {
-    return params.options._requester
-  }
-  return request
-}
-
 request.get = function (uri, options, callback) {
   var params = initParams(uri, options, callback)
   params.options.method = 'GET'
-  return requester(params)(params.uri || null, params.options, params.callback)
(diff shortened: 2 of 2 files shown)


nylen:
This is nice work, thank you for the PRs and for including tests :green_heart:  I'll nominate it for $30 bounty if you're interested.

I'd rather see us use `request` instead of `self` / `this`, do you think that would cause any problems?

tests/test-defaults.js
```diff
@@ -213,6 +213,35 @@ tape('recursive defaults', function(t) {
...
+      var headers = options.headers || {}
+      headers.baz = 'bar2'
+      options.headers = headers
+
+      defaultsOne(options, callback)
+    })
+
+  defaultsOne.get(s.url + '/get_recursive1', function (e, r, b) {
+    t.equal(e, null)
+    t.equal(b, 'TESTING!')
```
simov:
```js
t.deepEqual(r.request.headers, {foo: 'bar1'})
```

tests/test-defaults.js
```diff
@@ -213,6 +213,35 @@ tape('recursive defaults', function(t) {
...
+  })
+
+  defaultsTwo.get(s.url + '/get_recursive2', function (e, r, b) {
+    t.equal(e, null)
+    t.equal(b, 'TESTING!')
```
simov:
```js
t.deepEqual(r.request.headers, {baz: 'bar2', foo: 'bar1'})
```

tikotzky: That is being asserted over here: https://github.com/tikotzky/request/blob/fix-recursive-requester/tests/test-defaults.js#L72-L73

tests/test-defaults.js
tikotzky: That is being asserted over here: https://github.com/tikotzky/request/blob/fix-recursive-requester/tests/test-defaults.js#L65

tikotzky:
@nylen thanks for the offer of bounty nomination, but I'll pass on that for now.

As for changing to use `request` instead of `this`, are you referring to inside the `request.METHOD` function?

Changing that to use `request` instead of `this` would make it not use the `requester` function passed into `request.defaults`.

tests/test-defaults.js
simov: Right, you have to love that style of writing tests :) Anyway, not your fault.

tikotzky: yup I agree that is pretty disjointed, but thats not from this PR :)

nylen:
@tikotzky good point, :+1: from me then.  I'll merge this tomorrow if no one objects or beats me to it.

@simov @tikotzky our test suite has come a long way :grimacing: and further patches are always welcome.

tikotzky:
@nylen can this get merged now?

PR #1436, opened 2015-02-18
Add ability to set a requester without setting default options

tikotzky:
This PR adds the ability to call `request.defaults()` and only pass in a `requester` function.

Right now if you want to use defaults to just pass in a custom `requester` function you need to pass in an empty options object, like this...  
`request.defaults({}, function(uri, options, callback){})`

This PR adds the ability to do...  
```
request.defaults(function(uri, options, callback){})
// note we left out the first param of `{}`
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,12 @@ request.cookie = function (str) {
 }
 
 request.defaults = function (options, requester) {
+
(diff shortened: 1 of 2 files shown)


nylen:
:+1: thanks!

PR #1437, opened 2015-02-19
localAddress or proxy config is lost when redirecting

0x4139:
the localAddress or proxy configuration is lost when redirecting because the request object is initialized with an empty configuration instead of inheriting the one that made the initial request

### DIFF
--- a/tests/test-localAddress.js
+++ b/tests/test-localAddress.js
@@ -43,8 +43,6 @@ tape('bind to local address on redirect', function (t) {
         localAddress: localIPS[0]
     }, function (err, res) {
         t.equal(err, null)
-        res.body = "";
-        console.log(res.request.method);
         t.equal(res.request.localAddress, localIPS[0])
         t.end()
     })
--- a/tests/test-localAddress.js
+++ b/tests/test-localAddress.js
@@ -1,27 +1,51 @@
 'use strict'
-
 var request = require('../index')
-  , tape = require('tape')
+    , tape = require('tape')
 
-tape('bind to invalid address', function(t) {
-  request.get({
-    uri: 'http://www.google.com',
-    localAddress: '1.2.3.4'
-  }, function(err, res) {
-    t.notEqual(err, null)
-    t.equal(err.message, 'bind EADDRNOTAVAIL')
-    t.equal(res, undefined)
-    t.end()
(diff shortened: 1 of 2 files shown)


simov:
@0x4139 it would be great if you can provide a test case as well.

0x4139:
it seems that i can't bind on any interface in travis.

simov:
First of all this project uses 2 space indentation, can you check that in your editor? After that you can run `npm test` on your local workstation and fix all of lint errors, before pushing to GitHub.

nylen:
```js
'http://google.com', //redirects to 'http://google.com'
```

:confused: Maybe this should be `redirects to https://google.com`?

tests/test-localAddress.js
```diff
@@ -1,27 +1,49 @@
...
+        localInterfaces[ifname].forEach(function (iface) {
+            if ('IPv4' !== iface.family || iface.internal !== false) {
+                // skip over internal (i.e. 127.0.0.1) and non-ipv4 addresses
+                return;
+            }
+                localIPS.push(iface.address);
+        });
+    });
+    request.get({
+        uri: 'http://google.com', //redirects to 'http://google.com'
```
nylen: :confused: Maybe this should be `redirects to https://google.com`?

0x4139: fair enough, the problem is that travis doesn't have interfaces and the tests are failing :( , they are passing on my local machine but not on travis

simov: @0x4139 did you ever checked the travis-ci errors https://travis-ci.org/request/request/jobs/51369942 ?

nylen: or run `npm test` locally

0x4139:
@simov  sorry did not run npm test my bad.

0x4139:
Certificates are expired

simov:
@0x4139 can you pull https://github.com/simov/request/commit/1f6fbd57b8e2be0071e9771846790dda137ebb9a here? I removed the `proxy` parameter as there is a check for it inside the init method, and also added one for the `localAddress` After that this is a :+1:

simov:
Closing in favor of https://github.com/request/request/pull/1460

PR #1440, opened 2015-02-20
Add .editorconfig

simov:
Just to eliminate the indentation issues like in this PR here https://github.com/request/request/pull/1437

There is a plugin for pretty much every editor out there http://editorconfig.org/#download

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true


nylen:
I think I'm :-1: on this.  It's a good idea, but:

* people who don't pay attention to coding guidelines are unlikely to have such a plugin installed
* let's not have tons of `.files` in the project root.

simov:
Still we can't merge #1437 before the indentation is fixed and the eslint is passing https://travis-ci.org/request/request/jobs/51369941

nylen:
I agree, but I think that only certain programmers will have editors that honor `.editorconfig`, and those programmers are the same ones that will notice and follow our coding styles.  I personally use [ciaranm/detectindent](https://github.com/ciaranm/detectindent) with vim, which works pretty well most of the time.

simov:
Well we can easily tell them to install the editor config plugin for their IDE. Also I thought that the http://editorconfig.org/ approach is well established, but I can be wrong.

seanstrom:
@simov we can always add to the guidelines docs.
Perhaps a recommended tools section that lists some good tools for working on request.
Maybe the .editorConfig example isn't really important to list there.
But other tools for debugging and etc would be useful. What do you think?

seanstrom:
@simov we can always add to the guidelines docs.
Perhaps a recommended tools section that lists some good tools for working on request.
Maybe the .editorConfig example isn't really important to list there.
But other tools for debugging and etc would be useful. What do you think?

simov:
Yeah, this can be added, although the docs are yet to be re-organized, and people still have a hard time finding even the options section. Once the last two PRs are merged I'll bump the eslint version and enable the indent rule, so things like wrong indentation won't pass :)

PR #1442, opened 2015-02-20
More work to get the tests passing on node 0.12 & io.js

BBB:
I've fixed the issue with strictSSL tests by explicitly setting a cipher that matches the cert.

With node 0.12/ io.js there's now an error thrown by `socket#setTimeout` when it's passed a negative number. https://github.com/joyent/node/blob/master/lib/timers.js#L160

I've tried to catch this error when `req#setTimeout` is called here https://github.com/request/request/blob/master/request.js#L907 using a try catch, but it only seems to get caught in `process.on('uncaughtException')`. Any ideas how to handle this better; do we still need this test?


nicolasmccurdy:
Please merge master into your branch so we can see the results of the tests on Node 0.12 and io.js in CI (you also don't need the tests to run on 0.8, which should also be fixed when you update this with master).

simov:
Yes the test passes I'll push the msecs fix shortly and re-enable iojs/0.12 in the build

simov:
Thanks @BBB :+1:

BBB:
No worries @simov :)

PR #1443, opened 2015-02-20
Prevent crash with authorization and followRedirects of false

Cauldrath:
Currently turning followRedirects to false while passing in authorization information causes a crash.

self.redirects is only set if followRedirects or followAllRedirects is set:
https://github.com/request/request/blob/2f5f5e1913cad4b83fc379b1f96ac0f2f0fe1c7b/lib/redirect.js#L34

redirectTo is set to a value here if you are using authorization:
https://github.com/request/request/blob/2f5f5e1913cad4b83fc379b1f96ac0f2f0fe1c7b/lib/redirect.js#L66

I found this working with 2.27.0, but it seems to have been transferred to the latest code.

### DIFF
--- a/lib/redirect.js
+++ b/lib/redirect.js
@@ -105,11 +105,13 @@ Redirect.prototype.onResponse = function (response) {
     request._updateProtocol()
   }
 
-  self.redirects.push(
-    { statusCode : response.statusCode
-    , redirectUri: redirectTo
-    }
-  )
+  if (self.redirects) {
+    self.redirects.push(
+      { statusCode : response.statusCode
+      , redirectUri: redirectTo
+      }
+    )
+  }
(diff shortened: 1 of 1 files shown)


nylen:
Can we get a test that fails before this change?

Cauldrath:
Sorry, this was actually fixed already with this:
https://github.com/request/request/pull/1413/files#diff-ecf293ae0a8a5d19f25e88fc9e60363aR13

which isn't in the latest tagged version yet.

This is the test I was running:
```
var request = require('request');
request({
  uri: 'http://browserspy.dk/password-ok.php',
  followRedirect: false,
  auth: {
    user: "test",
    pass: "test",
    sendImmediately: false
  }
});
```

I also found a hack that gets around the problem on older versions:
```
var request = require('request').defaults({
  redirects: []
});
```

PR #1451, opened 2015-02-26
Reuse agentClass on redirect

### LINKED ISSUE #1304: POST 301 redirects don't respect "agent" property

mattcg:
Fixes #1304

As no options are passed to `init()` on redirect, Request fell back to the default agent. This fixes changes the behaviour so that the default is not used if `self.agentClass` is already specified.

Includes a test case.

### DIFF
--- a/request.js
+++ b/request.js
@@ -585,11 +585,11 @@ Request.prototype.init = function (options) {
     self.ca = options.ca
   }
 
-  if (!self.agent) {
-    if (options.agentOptions) {
-      self.agentOptions = options.agentOptions
-    }
+  if (!self.agent && options.agentOptions) {
+    self.agentOptions = options.agentOptions
+  }
 
+  if (!self.agent && !self.agentClass) {
     if (options.agentClass) {
       self.agentClass = options.agentClass
     } else if (options.forever) {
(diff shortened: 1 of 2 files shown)


polkovnikov-ph:
@mattch `not ok test-httpModule.js ........... 5/6`

mattcg:
Stupid, I was only checking test-redirect.js and not the whole suite.

I need help debugging this. Any idea why it might be happening? Is it related to pooling?

joshux:
I think I've found the problem: when redirect switches protocol, it'll call _updateProtocol, at [line 709 of redirect.js](https://github.com/mattcg/request/blob/721988b1638c7ef8292e5f3d5117ffb086ca1bac/request.js#L709) normally it will check if it's an http.Agent. But since we're using an custom agent, request couldn't know the agent's underlying protocol.

PR #1453, opened 2015-03-02
Test on Node.js 0.12 and io.js with allowed failures

nicolasmccurdy:
This is based on #1415 by @demohi, except I configured Travis to allow failures on Node 0.12 and io.js. This way, contributors can work on bug fixes and see the test results in CI without breaking the build for Node 0.10.


simov:
> This way, contributors can work on bug fixes and see the test results in CI

I can argue that anyone will ever see the failing tests (any of them) not only related to these ones currently failing on node 0.12 or io.js

The build says `All is well` with bold green, so unless I click on the _Details_ I'll never know what's going on.

nicolasmccurdy:
That's a good point, though I only planned on allowing failures until the 0.12 and io.js tests pass (while the fixes are in development). And even with the allowed failures, the build will still fail if 0.10 fails.

simov:
Isn't it better to have the failing status always? One additional click but that way you'll always check which one is passing. I just don't like the idea of silently passing on errors. Other than that I got the point and I agree that we should have 0.12 and iojs in the build script.

nicolasmccurdy:
That's a good point. Go ahead and merge your pull request if you'd like.

simov:
I think that's a better solution for now, than not having io.js and node 0.12 in the build.

PR #1458, opened 2015-03-04
Add cloneJar method

Narzerus:
Some complex libraries that depend on `request` require the ability to clone the cookie jar, because of the way the jar is implemented cloning a `tough-cookie` cookie jar will not work therefore the object must be re-built by setting each cookie individually.

I wanted to add some basic tests (though the code is tested in my end) but the way you guys are doing your tests is overcomplicated, maybe consider using mocha or another test runner. Also you can mock the `http` package with `nock`. But well thats a different topic

### DIFF
--- a/request.js
+++ b/request.js
@@ -282,7 +282,7 @@ Request.prototype.setupTunnel = function () {
   if (typeof self.proxy === 'string') {
     self.proxy = url.parse(self.proxy)
   }
-  
+
   if (!self.proxy || !self.tunnel) {
     return false
   }
@@ -298,7 +298,7 @@ Request.prototype.setupTunnel = function () {
   self.proxyHeaders = constructProxyHeaderWhiteList(self.headers, proxyHeaderWhiteList)
   self.proxyHeaders.host = constructProxyHost(self.uri)
   proxyHeaderExclusiveList.forEach(self.removeHeader, self)
- 
+
   // Set Agent from Tunnel Data
   var tunnelFn = getTunnelFn(self)
   var tunnelOptions = constructTunnelOptions(self)
@@ -1411,6 +1411,28 @@ Request.prototype.jar = function (jar) {
   }
   self._jar = jar
   return self
+}
+
+Request.prototype.cloneJar = function (cookieJar) {
+  var self = this
+  var newJar, cookieString, cookieStore, domains, domainPaths, cookieUrl
+
+  newJar = self.jar()
+  cookieStore = cookieJar._jar.store.idx
+  domains = Object.keys(cookieStore)
+
+  domains.forEach(function (domain) {
+    domainPaths = Object.keys(cookieStore[domain])
+
+    domainPaths.forEach(function (domainPath) {
+      cookieUrl = 'http://' + domain + domainPath
+      cookieString = cookieJar.getCookieString(cookieUrl)
+
+      newJar.setCookie(cookieString, cookieUrl)
+    })
+  })
+
+  return newJar
 }
 
 


seanstrom:
Is there a specific reason why you think Request would need to implement this?
Could you create the Jar separately and pass it into Request, then when you need to clone the jar you clone your reference.

```javascript
var j = request.jar()
var request = request.defaults({jar:j})

// later when i need to clone
var newJar = cloneJar(j)
```

In this example you would just have defined cloneJar locally.

Narzerus:
The reason is that cloning cookie jars is extremely important in some use cases, for example I'm building a scraping framework based on request, and I need to clone jars all the time to keep the session states under control.

They are basically useful as a session save point for more complex uses of the package.

@seanstrom deep cloning does not work in this case, I did try that but 
```javascript
  newJar !== j // returns false
```
also after cloning a lot of problems come up, this is probably because of private variables not being cloned.

seanstrom:
So Im starting to understand why cloning is important for your use case, but Im not seeing why it needs to be actually implemented inside of Request, aside from it being a minor convenience for your use case.
I still think you can accomplish what you want by wrapping Request.
Your `cloneJar` function doesn't seem to need any special internals of the Request instance, you only call `.jar()` to create a new Jar. It seems like this cloneJar function can be defined outside of Request and used with request instances.

Do you disagree?

Narzerus:
I do not think I'm the only one who needs to clone cookie jars, the reason why I'm submitting this PR is because anyone who wants to try and clone a "request flavored" jar will got into some nasty issues until they come up with rebuilding the jar completely.

This is why I disagree, since request is wrapping `tough-cookie`'s `CookieJar` there's no way of knowing how to clone a jar without reading the source code. Instead of doing that people should simply be able to call a .cloneJar() method. This is why I think it is request's responsibility to provide such method

seanstrom:
hmm i see what your meaning now, thanks for clarifying.
however we do need tests around this. Have you tried adding tests?

request.js
```diff
@@ -1413,6 +1413,28 @@ Request.prototype.jar = function (jar) {
...
+
+  newJar = self.jar()
+  cookieStore = cookieJar._jar.store.idx
+  domains = Object.keys(cookieStore)
+
+  domains.forEach(function (domain) {
+    domainPaths = Object.keys(cookieStore[domain])
+
+    domainPaths.forEach(function (domainPath) {
+      cookieUrl = 'http://' + domain + domainPath
```
seanstrom: should we be hardcoding `http`? Is it possible the cookie url would need `https`?

Narzerus: That was my only doubt, I tested it with https/http, thing is to my knowledge cookies only require domain, not protocol so this was added only because otherwise `getCookieString` wouldn't parse correctly.

Narzerus:
About tests, is there any guidelines to follow? Tests are extremely messy in this repo, no test runner or good assertion library, which would help a lot in the testing, so it looked really painful without some sort of guide.

I'm happy to do them with some guidelines though

request.js
seanstrom: I believe the protocol does matter. I would take a look at https://github.com/goinstant/tough-cookie . Seems you can set a cookie url with https. Also did you consider adding this to the `lib/cookies.js`

https://github.com/request/request/blob/a2c9ec7e28567bd58837b106abe1b9a03d881e1d/lib/cookies.js

seanstrom:
Summoning @nylen @simov to chime in

lalitkapoor:
@Narzerus from the package.json I see that we're using taper to run the tests. so it's `node node_modules/.bin/taper tests/test-*.js` to run them all

nicolasmccurdy:
You can also run the whole test suite (which includes some extra stuff) with `npm test`.

Narzerus:
@lalitkapoor I'll have a look and create the tests then.

@seanstrom about protocol, I'll have a closer look though since this is only retrieving the cookie string from tough-cookie it should only matter when using the `getCookie` method which we are not. Cookies have a 'Secure' property which is present on https protocol. Since the cookie is already set, if it was set via https it should come (will test this) with that option already because otherwise cookieStore would index with protocol included, but it doesnt.

Also I did first consider adding the method at `lib/cookies.js`, then I realized that that module is only exporting the `.jar()` method so it made no sense.

seanstrom:
Well the module actually exports a `parse` method as well, I don't see why it couldn't be moved there either way.

request.js
```diff
@@ -1413,6 +1413,28 @@ Request.prototype.jar = function (jar) {
   return self
 }
 
+Request.prototype.cloneJar = function (cookieJar) {
+  var self = this
+  var newJar, cookieString, cookieStore, domains, domainPaths, cookieUrl
+
+  newJar = self.jar()
+  cookieStore = cookieJar._jar.store.idx
```
seanstrom: Looking more into this it seems that relying on `idx` is brittle.
`request.jar` takes in a store argument, and that gets passed all the way down to tough-cookie's CookieJar.
That doesn't rely on `idx` directly, but an API on the store. Meaning that it's not required to have the property `idx` it could be named anything.

seanstrom: Long story short, we need a better way to get to iterate over the jar's store

seanstrom: Looking into this, there isn't a direct way to get the information we need.
Let me know if you find anything different.

Narzerus: Let me try some stuff, I'll let you know, I didn't want to iterate over the idx keys either but so far it has proven the only useful way. Passing a store to the request.jar() method gives a tough-cookie error at some point where the store is asyncronous.

seanstrom: Yeah tough-cookie requires that the store has a synchronous interface I believe

lalitkapoor: Stores can be async or sync, however request requires that the underlying store is synchronous.

Narzerus: Yup, so basically the problem is that you can't just pass a `request` jar's store and create a new tough-cookie Cookie Jar, because now it lacks of the methods that `request` requires. For example, `request` adds a `getCookieString() ` method which is really using `tough-cookie`'s `getCookieStringSync()`, which will cause an error at some point in tough-cookie's end.

My PR is building the cookie jar correctly, I agree though I should make sure that the way I'm **retieving** the cookies is the best way. tough-cookie's public API's documentation doesn't look

seanstrom: Seems like we want a supported way to get all the cookies from the cookie store without reaching into "private" interfaces. We may want to discuss this with tough-cookie directly.

request.js
```diff
@@ -1413,6 +1413,28 @@ Request.prototype.jar = function (jar) {
   return self
 }
 
+Request.prototype.cloneJar = function (cookieJar) {
+  var self = this
+  var newJar, cookieString, cookieStore, domains, domainPaths, cookieUrl
+
+  newJar = self.jar()
```
seanstrom: I don't think this works. I'm running your code on my end and Im not getting a Jar returned from this statement.
Im getting back the request instance. Seems like Request.prototype.jar takes in a jar as well.
Perhaps you meant to say `cookies.jar()` ?

Narzerus: You are right, one sec let me check what's wrong, I though I was calling request's public method jar() which returns a new jar.

Narzerus: I'm gonna add the tests and then fix this so that I make sure this is done correctly. I'll let you know

FredKSchott:

PR #1459, opened 2015-03-05
Add option to time request/response cycle (including rollup of redirects)

aaron-em:
Sure, it'd be fairly trivial to grab timestamps before initiating the request, and in the response callback, and compare them; on the other hand, I'll eventually have a _lot_ of tests where I want to use the functionality, and writing a wrapper around the request library doesn't appeal to me.

### DIFF
--- a/request.js
+++ b/request.js
@@ -874,7 +874,6 @@ Request.prototype.start = function () {
   // start() is called once we are ready to send the outgoing HTTP request.
   // this is usually called on the first write(), end() or on nextTick()
   var self = this
-    , startTime
 
   if (self._aborted) {
     return
@@ -898,11 +897,10 @@ Request.prototype.start = function () {
 
   debug('make request', self.uri.href)
 
-  startTime = new Date().getTime()
   self.req = self.httpModule.request(reqOptions)
 
   if (self.timing) {
-    self.startTime = startTime
+    self.startTime = new Date().getTime()
   }
 
   if (self.timeout && !self.timeoutTimer) {
--- a/request.js
+++ b/request.js
@@ -282,7 +282,7 @@ Request.prototype.setupTunnel = function () {
   if (typeof self.proxy === 'string') {
     self.proxy = url.parse(self.proxy)
   }
-  
+
   if (!self.proxy || !self.tunnel) {
     return false
   }
@@ -298,7 +298,7 @@ Request.prototype.setupTunnel = function () {
   self.proxyHeaders = constructProxyHeaderWhiteList(self.headers, proxyHeaderWhiteList)
   self.proxyHeaders.host = constructProxyHost(self.uri)
   proxyHeaderExclusiveList.forEach(self.removeHeader, self)
- 
+
   // Set Agent from Tunnel Data
   var tunnelFn = getTunnelFn(self)
   var tunnelOptions = constructTunnelOptions(self)
@@ -546,6 +546,11 @@ Request.prototype.init = function (options) {
   }
   if (options.multipart) {
     self.multipart(options.multipart)
+  }
+
+  if (options.time) {
+    self.timing = true
+    self.elapsedTime = self.elapsedTime || 0
   }
 
   if (self.body) {
@@ -869,6 +874,7 @@ Request.prototype.start = function () {
   // start() is called once we are ready to send the outgoing HTTP request.
   // this is usually called on the first write(), end() or on nextTick()
   var self = this
+    , startTime
 
   if (self._aborted) {
     return
@@ -891,7 +897,13 @@ Request.prototype.start = function () {
   delete reqOptions.auth
 
   debug('make request', self.uri.href)
+
+  startTime = new Date().getTime()
   self.req = self.httpModule.request(reqOptions)
+
+  if (self.timing) {
+    self.startTime = startTime
+  }
 
   if (self.timeout && !self.timeoutTimer) {
     self.timeoutTimer = setTimeout(function () {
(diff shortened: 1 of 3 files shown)


nicolasmccurdy:
:+1:

aaron-em:
@nicolasmccurdy Good catch, thanks! Odd that that assertion passes locally but fails in CI - in any case, the check around the mock delay on redirect renders the failing assertion moot, so I just removed it.

nicolasmccurdy:
:+1:

seanstrom:
The original tests sometimes fail for me.
I notice the failing spec is testing that the non-redirected request is faster than a request that is redirected. Was that the intent?

seanstrom:
If you still want that test you would probably need to increase the `redirect_mock_time` to at least 20ms.
I think 10ms is too low for `setTimeout`, possible the V8 isn't accurate enough for 10ms?

aaron-em:
I doubt it's a V8 issue since the test was reliable locally, and only failed in CI. 

--
Aaron Miller
Mobile: (443) 810-0023
me@aaron-miller.me
http://aaron-miller.me

aaron-em:
Not really; the intent was to ensure that, when following redirects, the time reported is the sum of all request times, instead of only the time taken by the last request made. 

--
Aaron Miller
Mobile: (443) 810-0023
me@aaron-miller.me
http://aaron-miller.me

seanstrom:
What I was getting at is that running your original test locally for me was not reliable, it would fail or succeed on and off.

seanstrom:
The original test was just checking the redirect elapsed time was greater than the non redirect elapsed time right? Not the sum of each request's elapsed time correct?

seanstrom:
I've also have had different results depending on the order of the tests.

request.js
```diff
@@ -891,8 +897,14 @@ Request.prototype.start = function () {
   delete reqOptions.auth
 
   debug('make request', self.uri.href)
+
+  startTime = new Date().getTime()
   self.req = self.httpModule.request(reqOptions)
 
+  if (self.timing) {
+    self.startTime = startTime
```
seanstrom: Could we just define `self.startTime` to be `new Date().getTime()` ?

tests/test-timing.js
```diff
@@ -0,0 +1,62 @@
...
+    plain_server.on('/redir', function (req, res) {
+      // fake redirect delay to ensure strong signal for rollup check
+      setTimeout(function() {
+        res.writeHead(301, { 'location': 'http://localhost:' + plain_server.port + '/' })
+        res.end()
+      }, redirect_mock_time)
+    })
+
+    request('http://localhost:' + plain_server.port + '/', {}, function(err, res, body) {
+      t.equal(err, null)
```
seanstrom: why do we have this test here?

tests/test-timing.js
```diff
@@ -0,0 +1,62 @@
...
+    })
+
+    t.end()
+  })
+})
+
+tape('no-op', function(t) {
```
seanstrom: any reason we have this test?

seanstrom:
@nylen @simov @FredKSchott Do you guys have any opinion about adding this?
I'm thinking it would be a good idea.

tests/test-timing.js
aaron-em: to ensure that, when redirects are handled inline, the time reported on the response is the sum of all request times, rather than just the time taken by the last request/response cycle

seanstrom: Im confused, you comment seems to be explaining a different test.
This test is expecting that `err` is equal to `null`.

seanstrom:
Seems like the tests are failing because an error that was thrown by Karma Spec Runner

aaron-em:
@seanstrom this is my first experience with travis; i have no idea what to do about that, sorry.

nicolasmccurdy:
It seems to me like this might be an issue with Karma occasionally failing on Travis. Could someone please restart the job so we can see if that helps?

seanstrom:
That's okay, i just wanted that info to stated on the PR.
That way it doesn't seem like your tests were failing or Request's tests were failing.

seanstrom:
I attempted to restart the build, and Travis says the job has been restarted.
But I get no visualization of the restart being in process. Does anyone else know how Travis restart works?

seanstrom:
Okay so closing and opening the PR will restart the PR build in Travis CI

seanstrom:
@aaron-em okay everything looks good, but I made one other line comment.
After that's addressed we should be good.

tests/test-timing.js
aaron-em: Wups, you're right - that's what I get for replying from my phone.

simov:
@seanstrom I don't think this should be part of the core as it doesn't add any value to the module. You can wrap request in a function and achieve the exact same result without passing additional option to request itself.

But if you want it merge it, it's your call :)

simov:
Actually now when I think about it, why not, we have tons of other options as well. So @seanstrom I'm not against adding it.

seanstrom:
@simov yeah that was my feeling at first, but like you said we have a bunch of useful options.
I could see it being a useful utility to have around.

seanstrom:
@aaron-em any other changes you want to make?
Im ready to merge this unless anyone else has objections

aaron-em:
I've done all the damage I mean to do; feel free to merge, and thanks!

--
Aaron Miller
Mobile: (443) 810-0023
me@aaron-miller.me
http://aaron-miller.me

seanstrom:
@simov do you have access to publish the module or is that just @nylen ?

simov:
@seanstrom check out the names listed below the maintainers section here https://www.npmjs.com/package/request

pungoyal:
when do we expect an npm release for this? it would be very helpful for us.

simov:
The next release on NPM is scheduled for this week, I'll post an issue in the tracker when it's done :+1:

pungoyal:
:+1: look forward

Issue #1421, opened 2015-02-12
Fix request with Node.js 0.12.x ($100)


nylen:
I'm nominating this fix for a $100 bounty.  Fellow contributors, what do you think?

mikeal:
Is request broken or are the tests just having issues? I fixed all but one test on io.js 1.2 and the fixes were entirely in the tests.

nylen:
I've been seeing some `Error: socket hang up` and `ECONNRESET`, like what's in the HTTPS tests.  Haven't had time to look into it.

akshayp:
@nylen @mikeal I haven't been able to dig into this deeply yet but it looks like the ECONNRESET errors happen for me when the forever agent is being used. I'll post back with a reproducible test case when I have one

rvagg:
If you want a reproducible test case you could look at the test suite in [st](https://github.com/isaacs/st) which I can't make pass on 0.12 or io.js because of this. Run `node test/multi-mount.js` and you'll see it.

@othiym23 mentioned that there was a problem with npm and io.js that I have a feeling may be related to this, pulling him in here to confirm or deny.

rvagg:
Actually .. I think this is simply just the new `maxSockets` so users need to be told to manage that themselves now that it's `Infinity` instead of `5`.

Try:

```
const http = require('http');
const port = 1337;
const clients = 500;

var done = 0;
var agent = http.Agent()
// agent = http.Agent({ maxSockets: 100 })

var server = http.createServer(function (req, res) {
  res.write('.');
  console.log('waiting 100 @ %d', done)
  setTimeout(function () {
    console.error('ended')
    res.end('.');
  }, 100);
}).listen(port, function () {
  for (var i = 0; i < clients; i++) {
    (function (i) {
      console.log('requesting @ %d', i)
      http.get({ port: port, agent: agent }, function (req) {
        req.resume()
        req.on('end', function () {
          console.log('done %d', done - 1)
          if (++done === clients)
            process.exit(0);
        });
      })
    }(i))
  }
})
```

akshayp:
@rvagg  In my case I'm using the forever option which never works presumably because the HTTP Agent API changed quite a bit with node 0.12. eg:

```js
var request = require('./index').forever({maxSockets: 1});
request('http://www.yahoo.com', function (error, response, body) {
    if (error) {
        console.log(error);
    } else {
        console.log(response.statusCode);
    }
});
```
 leads to

```
[...]

Issue #1414, opened 2015-02-07
Is request support with nodejs 0.12


stableShip:
update nodejs to 0.12 from 0.10.36 .  request -v: 2.53.0 . This Error occur:
![image](https://cloud.githubusercontent.com/assets/2979404/6091684/1338b580-aef3-11e4-837f-171bcd271b0d.png)

timmckenzie:
Confirmaiton, all tests passed in node 0.10.36, but failed in 0.12.0.  Running on OSX.

```
ok test-agentOptions.js ............. 4/4
ok test-basic-auth.js ............. 19/19
ok test-bearer-auth.js ............ 14/14
ok test-body.js ................... 13/13
ok test-cookies.js ................ 17/17
ok test-defaults.js ............... 23/23
ok test-digest-auth.js ............ 11/11
ok test-emptyBody.js ................ 9/9
ok test-errors.js ................... 8/8
ok test-event-forwarding.js ......... 4/4
ok test-follow-all-303.js ........... 3/3
ok test-follow-all.js ............... 3/3
ok test-form-data-error.js .......... 1/1
ok test-form-data.js .............. 40/40
ok test-form-urlencoded.js ........ 12/12
ok test-form.js ................... 18/18
ok test-gzip.js ................... 17/17
ok test-hawk.js ..................... 3/3
not ok test-headers.js ............ 20/22
ok test-http-signature.js ........... 4/4
ok test-httpModule.js ............. 50/50
not ok test-https.js .............. 10/20
ok test-isUrl.js .................. 15/15
ok test-json-request.js ........... 27/27
ok test-localAddress.js ............. 5/5
ok test-multipart.js .......... 1136/1136
ok test-node-debug.js ............. 20/20
ok test-oauth.js .................. 47/47
ok test-onelineproxy.js ............. 3/3
ok test-option-reuse.js ........... 10/10
ok test-params.js ................. 12/12
ok test-piped-redirect.js ........... 2/2
ok test-pipes.js .................. 27/27
ok test-pool.js ..................... 4/4
not ok test-proxy-connect.js ........ 4/5
ok test-proxy.js ................ 104/104
ok test-qs.js ..................... 16/16
ok test-redirect-auth.js .......... 18/18
ok test-redirect-complex.js ....... 40/40
ok test-redirect.js ............... 67/67
ok test-rfc3986.js ................ 22/22
ok test-timeout.js .................. 9/9
```

jspavlick:
[...]

PR #1468, opened 2015-03-09
Re-enable io.js/node 0.12 build

simov:
Fixed the negative msecs error and re-enabled io.js and node 0.12 in the build.

Closes https://github.com/request/request/issues/1421
Closes https://github.com/request/request/issues/1414


package.json
```diff
@@ -41,7 +41,7 @@
     "isstream": "~0.1.1"
   },
   "scripts": {
-    "test": "npm run lint && node node_modules/.bin/taper tests/test-*.js && npm run test-browser && npm run clean",
+    "test": "npm run lint && node node_modules/.bin/taper tests/test-*.js && npm run test-browser",
```
nicolasmccurdy: Out of curiosity, why was this removed?

nicolasmccurdy:
:+1: Nice job!

package.json
simov: I think we had a clean script back then, that logic was later moved to the browser test itself. NPM on node 0.10 doesn't complain about the missing script, where the newer version of it throws an error.

nicolasmccurdy: Ah, it seems like this is safe to remove then. npm won't look anywhere else for scripts to run (see [run](https://docs.npmjs.com/cli/run-script)).

eiriksm: Yes. I think the clean script removed a browserified build of the test script, but the current version browserifies the test script on the fly instead.

Issue #1473, opened 2015-03-10
Array parameters in the query string produce numeric indexes


gajus:
```js
request({uri: 'http://foo.com', qs: {filter: ['a', 'b']}});
```

Will make a request to http://foo.com?filter[0]=a&filter[1]=b. It should be making a request to http://foo.com?filter[]=a&filter[]=b.

simov:
Hi @gajus did you tried the `useQuerystring` option?

> useQuerystring - If true, use querystring to stringify and parse querystrings, otherwise use qs (default: false). Set this option to true if you need arrays to be serialized as foo=bar&foo=baz instead of the default foo[0]=bar&foo[1]=baz.

gajus:
`useQuerystring` is even worse, because it does not produce `[]` as part of the name.

simov:
It seems that this issue was recently addressed in the qs module https://github.com/hapijs/qs/pull/70

So the syntax would be (currently undocumented, scheduled for version 2.4.0)

```bash
> qs.stringify({arr:['a', 'b']}, {})
'arr%5B0%5D=a&arr%5B1%5D=b'
> qs.stringify({arr:['a', 'b']}, {arrayFormat: 'indices'})
'arr%5B0%5D=a&arr%5B1%5D=b'
> qs.stringify({arr:['a', 'b']}, {arrayFormat: 'brackets'})
'arr%5B%5D=a&arr%5B%5D=b'
> qs.stringify({arr:['a', 'b']}, {arrayFormat: 'repeat'})
'arr=a&arr=b'
```

Once it's published we should improve the `qs` option key to be able to receive additional options (just like the `multipart` key) As for the chain API it won't be a problem I guess.

Lets keep this issue open for now.

simov:
`qs@2.4.0` just got published, assigning this one to myself, I'm going to introduce a `qsOptions` option, I think it will be cleaner in our case.

PR #1477, opened 2015-03-13
Add support for qs options via qsOptions key

simov:
Closes https://github.com/request/request/issues/1473

The `qsOptions` key can be used to pass options to both the `parse` and the `stringify` methods of the `qs` and `querystring` modules.

The `arrayFormat` options was missing from the `qs` readme https://github.com/hapijs/qs/pull/72

### DIFF
--- a/README.md
+++ b/README.md
@@ -565,6 +565,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 
 * `uri` || `url` - fully qualified uri or a parsed url object from `url.parse()`
 * `qs` - object containing querystring values to be appended to the `uri`
+* `qsOptions` - object containing options to pass to the `qs` or `querystring` module
 * `useQuerystring` - If true, use `querystring` to stringify and parse
   querystrings, otherwise use `qs` (default: `false`).  Set this option to
   `true` if you need arrays to be serialized as `foo=bar&foo=baz` instead of the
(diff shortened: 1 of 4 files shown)


nylen:
A few thoughts on this:

1. Can we link directly to the relevant options from the readme?
  - [`qs.parse`](https://github.com/hapijs/qs#parsing-arrays)
  - [`qs.stringify`](https://github.com/hapijs/qs#stringifying)
  - [`querystring.parse`](https://nodejs.org/docs/v0.12.0/api/querystring.html#querystring_querystring_parse_str_sep_eq_options) (seems better to link directly to a specific Node.js version here, rather than the latest)
  - [`querystring.stringify`](https://nodejs.org/docs/v0.12.0/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options)
- Not our issue, but it's hard to tell from a glance what options the `qs` methods support.
- Since this does a lot of different things, may be better to split it into two: `qsParseOptions` and `qsStringifyOptions`?
- Some text like this might be good, too:

> `qsOptions` - object containing options to pass to the `qs` or `querystring` module*, for example, to change the way arrays are converted to query strings*

simov:
Agreed, I was thinking about the same problem - different options for `parse` and `stringify` Also adding links there would be good.

simov:
Improved the docs a bit, and introduced the `qsParseOptions` and the `qsStringifyOptions` keys.

PR #1481, opened 2015-03-14
Fix baseUrl and redirections.

burningtree:
I found a bug in new `baseUrl` feature, when used in combination with redirection.

This script produce absurdly error `[Error: options.uri must be a string when using options.baseUrl]`:

```javascript
var request = require('./node_modules/request');
var http = require('http');

var s = http.createServer(function(req, res) {
  if(req.url === '/redirect/') {
    res.writeHead(302, {
      location : '/'
    })
  } else  {
    res.statusCode = 200
  }
  res.end()
}).listen(6767);

request({
  url: '/',
  baseUrl: 'http://localhost:6767/redirect'

}, function (err, resp, body) {
  console.log(err);
  s.close();
});
```
This is due to the fact that first call changes URI to an object, and therefore in the second iteration (to redirected page) can be seen as a string. The solution is delete `baseUrl` immediately after use, because certainly no longer needed.

This pull request fixes this bug. Test included.

### DIFF
--- a/request.js
+++ b/request.js
@@ -386,6 +386,7 @@ Request.prototype.init = function (options) {
     } else {
       self.uri = self.baseUrl + '/' + self.uri
     }
+    delete self.baseUrl
   }
 
   // A URI is needed by this point, throw if we haven't been able to get one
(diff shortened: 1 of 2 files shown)


simov:
Good catch @burningtree :+1:

PR #1483, opened 2015-03-15
fix util is not defined

AndrewTsao:


simov:
Hi, @AndrewTsao , branch 3.0 isn't touched for more than a year and it's effectively a dead one. You can check the current branch state here https://github.com/request/request/branches

mikeal:
ya, 3.0 is sorta dead. to reboot that effort you'd need to pull the code out and re-branch. the base is solid and is the only streams2 based duplex request.

Issue #1472, opened 2015-03-10
forever-agent does not work with Node 0.12


akshayp:
Moving https://github.com/request/request/issues/1421#issuecomment-74371413 since this technically doesn't block request from working unless the forever option is enabled

I'm using the forever option with request. Starting 0.12 this option leads to ENOTSOCK and ECONNRESET errors. This never works presumably because the HTTP Agent API changed quite a bit with node 0.12. eg:

```js
var request = require('./index').forever({maxSockets: 1});
request('http://www.yahoo.com', function (error, response, body) {
    if (error) {
        console.log(error);
    } else {
        console.log(response.statusCode);
    }
});
```
 leads to

```
{ [Error: connect ENOTSOCK] code: 'ENOTSOCK', errno: 'ENOTSOCK', syscall: 'connect' }
```

vs on node 0.10 it logs

```
200
```

nicolasmccurdy:
It seems like it would help to write a test for this functionality.

akshayp:
@nicolasmccurdy Added a test in the linked PR

PR #1486, opened 2015-03-15
Add a test for the forever agent

akshayp:
@nicolasmccurdy Adding a test for https://github.com/request/request/issues/1472. This passes on node 0.10 and fails on 0.12.x and iojs

Closes https://github.com/request/request/issues/1472

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "dependencies": {
     "bl": "~0.9.0",
     "caseless": "~0.9.0",
-    "forever-agent": "~0.5.0",
+    "forever-agent": "~0.6.0",
     "form-data": "~0.2.0",
     "json-stringify-safe": "~5.0.0",
     "mime-types": "~2.0.1",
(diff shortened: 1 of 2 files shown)


nicolasmccurdy:
:+1:

FredKSchott:
+1 This should probably be merged, even if it is breaking. Any objections @nylen @seanstrom @simov ?

seanstrom:
I agree we should take this, but we need to patch asap.
Im willing to take a look at this tonight/tomorrow.

simov:
@seanstrom that would be great. Although having this PR and the failing test is good, I don't think merging it without a fix is the right thing to do, as it may stay that way indefinitely.

seanstrom:
Okay good news, I got the tests to pass by patching ForeverAgent.
I'll have a PR for that up really soon.

seanstrom:
Here's the PR for the patch, https://github.com/request/forever-agent/pull/25

simov:
@akshayp can you bump the `forever-agent` version to `0.6.0` and push so that we can see the tests passing.

akshayp:
@simov Pushed and tests are now all passing. Thanks for the patch @seanstrom!

PR #1493, opened 2015-03-19
Support/refactor init method

seanstrom:
Hello everybody,

it's been a while since I've submitted a refactoring PR, and this time I decided to try to break up and clean the logic that's happening in the Request.prototype.init function.
At the moment I've refactored a bunch of code into some well-meaning functions.
I also leveraged the `constructObject` function heavily, I do notice that's not necessary in every place I use it, but the I do feel that it's useful in most of the areas.

The biggest difference is that the init logic has been moved to another module.
This is more of a temporary refactor, there was just to much logic polluting the request.js file.
Another big difference is the way I'm doing some error handling, you'll see what I mean, but just know I'm not completely set on that implementation.

Overall this is still a work in progress, but I would like to hear some feedback.

Summoning: @simov @nylen @FredKSchott @mikeal


simov:
OK, I was about to be a bit harsh as usual, but generally I like the idea and I want to give this PR a chance, though there are a couple of striking issues that I want to see resolved first, before I give it another read.

1. `lib/request/init.js` - why is this in separate folder? Put it directly into `lib`
2. `constructObject` is really gross, it was not used in request.js at all! Just forget about it, really. I was about to show you examples on what I mean, but I'm pretty sure you know how to refactor all of the places to use a simple assignment operator.
3. No need for `props` in `handleBaseUrl` and probably the rest of the places too. You can just return `new Error` after that in the consumer you are checking for error anyway, so you can check for `instanceof Error` instead
4. Use consistent spacing between `if(` | `function (` | `function urlToUri(` with the rest of the code base. I'm planning on updating `eslint` to the latest version, but in the mean time lets keep consistent with what we have atm. After that we can have a heated discussion on which `eslint` options should be enabled or not.
5. Commented out test - refactoring should not change application logic!
6. `// Dirty Function` - I really don't like silly comments in code

And of course there are a couple of things that I do like in your refactoring, so it's not all bad :) We should have a few more iterations on this, but for now this PR is a no go for me. I hope you find my comments constructive, @seanstrom

seanstrom:
1. I put that file in a request folder in order to namespace the file. Ill just change that request-init.js and move it up a directory.
2. constructObject is just another helper in the helper.js I see no reason for it to not be used elsewhere in the code base. As far as it being gross I would disagree, but I am partially biased.
3. Props is just an implementation detail, the main idea here was that we're not mutating self directly all over the place. So when I refactored logic out to functions i didn't want to pass self if I didn't need to.
In the end I chose to return an object that has an error field, meaning that I don't need to check if it is of type Error or not.
4. The codebase has been inconsistent about whitespace for a while, but I'll tidy up my changes.
5. I had a good reason for commenting out this test. I believe that what this tests is proving isn't really important, before we added the BaseUrl logic, we would gladly accept a URI object as the first argument.
With the addition of the base url code it forced users to not pass that in as the first argument, which I believe to be a change in expected behavior. The reason any of this is important is because we should parse the URI at the beginning of the `init` method instead of mangling a string part of the way and then finally midway parsing it to some object, and then have everyone assume its an object afterwards.
6. That was for reference, forgot to remove it. It was just a marker on function that mutates `this` directly and returns no values.

I personally think that is PR is half finished work, there's plenty to be done and cleaned up and then conventionalized.

simov:
For `constructObject` I'll put it that way - you are introducing abstraction in places where it's not needed - you are not copying nor extending anything. The code was perfectly fine without that _cool_ wrapper. Just take a look at the code how it looks too, where it should be just a simple assignment you have tons of nonsense like extend/done all over the place. If not doing it for yourself think about it that way - you shouldn't expect from our contributors to have deeper understanding of javascript or some weird parts of our code base, unless you want to carry the burden of maintaining it alone, or just with a few hand picked people that might, or might not be around for a while.

For `props` - same thing. Unneeded layer of abstraction, much less dangerous but still adds some clutter, because you are checking for errors in the consumer of these functions anyway - do not assume that `instanceof` should be slower than `obj.property` check.

For `baseUrl` reasoning see this https://github.com/request/request/pull/1469#issuecomment-78073749

seanstrom:
The argument against `constructObject` is more of a knee jerk reaction of, "That's unusual".
And on top of that I haven't defended the use of `constructObject` in all of these areas I have used it.
I know it may not be appropriate to use it even for it's very purpose of existing and I even gave heads up to any/all reviewers that I wasn't set on the implementation.

As far as other contributors, I think expecting contributors to read the code base is a given.
Plus, this isn't black magic javascript, its just a simple function.
Whether it be needless abstraction is an opinion and nothing more.

As for `props`, that isn't even anything clever.
It's a simple object, I return an object in some functions. That's just apart of its signature.
I could have made it just return the Errors, but I wanted to make it more open ended for other types of Errors to be introduced without needing to add more instanceof checks.
I perhaps could also just assume that anything truthy that's returned from the function is an error.
Also I never mentioned anything about performance with `instanceof` vs `obj.prop`.

As for the reasoning about having the baseUrl throw if the uri is not a string, I'm not sure any explanation is given, just that you said to do that. It seems weird that baseUrl is throwing when uri isn't a string.
There's several places in the code that already emit/throw errors for uri isn't given. I think baseUrl should work off a URI object instead.
In any case Ill revert the commented out test, I'll patch the code to be backwards compatible since that's the proper thing to do.

seanstrom:
I would also note that it would be probably more effective to ask me why I wrote the code in a certain way. There could be knowledge or thought process or context behind the way it's written.
This way we can share the same context and make decisions together, instead of on our own.

simov:
@seanstrom I told you in my first comment, I'm seeing the good sides of your refactoring work too. I'm talking only about the negative sides, and I'm trying to argument myself as good as possible.

seanstrom:
Im not saying that you haven't given me enough praise, im just explaining that the reason I used `constructObject` and the `props` objects were for a good reason.
The goal being that we when we extract logic to functions we don't just mutate `self` directly but provide data for self to assign props to in the init function. I just abstract over the assignment process by using `constructObject` but that's really an implementation detail and not preference.

seanstrom:
Also sorry if i'm being pushy too, I forget that things over the internet always read a bit harsher than they're meant. So I too know that I may not be very effective by just arguing.
And on that note, I've pushed up several clean up updates.
Ones on primarily removing some `constructObject` code that was doing more bloating of lines than really building objects. Check it out.

seanstrom:
Oh its a cruel day. Karma has decided to blow during test runs... :cry:

FredKSchott:
+1 to the idea, I always love refactoring & cleanup

I've said this in a few different places, but I'm against pulling something significant out of a function just to save lines of code. Conceptually, init is a good chunk of work on/belonging to request so to pull it out entirely into a separate file would be incorrect under that line of thinking. Additionally, I'll now have to jump between request.js and request-init.js and am more likely to miss something / less likely to understand what is going on.

So I love all the work and thought put into this, but I'd like to see the core "init request" logic remain in request.js, either in the request constructor or on a request.init() method.

FredKSchott:
(Having said that, I've been so busy with new job lately that I've been completely taken away from this project, so feel free to merge/not merge without me. Just wanted to pop in and add my usual "removing lines of code != removing complexity" reminder :)

seanstrom:
@FredKSchott Im actually more on board with your line of thinking, given the current implementation.
I actually did most of my work inside `request.js` but at the end moved things out to another file.
I did that so it would be easier to refactor, but I hope to be able to move the new refactored function back. At the moment I'm trying to big movements of code, and I feel I can do that with more confidence when the refactoring is happening in another file. I thought of it as a sanitary environment for my surgery.

simov:
Thanks @seanstrom I will be able to review very early in the morning your time :) Also I think what happens in this thread is good for the project, though it might not be the most comfortable :)

@FredKSchott the same argument for jumping between files, can be applied to scrolling hundred of lines back and forth. I personally think what @seanstrom is doing with this PR is a good. Also I'm for keeping the init logic separated.

FredKSchott:
@seanstrom haha I like the surgery analogy, sounds good :)
@simov It's definitely subjective, but my main point was that removing logic blocks of code from a giant file (helpers, core modular units of request funcitonality, etc) is 1000x more effective than just moving around code for the sake of line numbers.

nylen:
@seanstrom you & I have discussed similar issues before.  Don't pull too many things into functions, don't create cute/clever helpers.  Please listen to @simov's feedback instead of arguing with him, he's right.

seanstrom:
@nylen, @simov aren't discussing method extraction.
We were discussing helpers, and i think saying they're cute/clever isn't a bad thing.
Those cute/clever helpers can also be useful. Deciding whether they're useful or not is a matter of opinion. Saying you're right, or they're right, or I'm right based on preference or bias is wrong.
I would rather they're be constructive feedback and strong arguments, aside from one's that come off as, "That's just what you do!". When someone gives off that vibe, it makes your opinion look like law, which it isn't.

Furthermore I refuse to be told, "Don't talk back and do what we say" as an reason to implement code a certain way. I've been pretty reasonable in the past to changes my code to match the ideals of the project. At this point I'm only suggesting alternative implementations, which aren't broken.
They just seem to strike a cord with some people in the JS community as to much abstraction.
I disagree with that mindset, I chose to explore a possible new way of doing things, while having good reasons for doing so. I'm trying to bring these alternatives to the project, and ask the question "Why not this?". Even though the conversation started off a little rough, doesn't mean @simov and I are beyond reasoning with each other as peers.

froatsnook:
@seanstrom The problem with using a parsed URI for baseUrl is that the semantics become unclear.  As it is now, you give a string baseUrl and a string uri.  The two are concatenated and that's your URL.  When either/both can be parsed URIs, then what happens with query parameters and anchors in the baseUrl?  Are they replaced by the uri?  Or added?  It's just not a useful enough feature to warrant a detailed explanation in the README when we can restrict the types to strings and concatenate them and we're done.  That's easy for users to understand.

I guess the alternative would be to throw on query parameters or anchor in the parsed URI baseUrl, and throw on scheme or host or auth in the uri.  I started out with an implementation like this, but again, it was looking very complicated, going through all the potential combinations, for such a tiny feature.

seanstrom:
@froatsnook What I had implemented was URI was always an object and baseUrl was still a string.
At the moment I've worked around the baseUrl feature to still work with strings only.

Im not seeing the whole problem clearly at the moment, so I won't try to make any comments or changes until I think of some good questions/possible solutions.

simov:
@seanstrom  there are three `constructObject` references left. That's the first thing, honestly I have a hard time following all the props, done, and the tons of state variable assignments back and forth. This PR is still looking far from ready for merging to me, what I outlined yesterday were the first striking issues that I saw first, and I'm not seeing even them fully resolved.

So lets make it that way - once you remove the `constructObject` object wrapper, I can play around with your code, just to be a bit more comfortable with my further suggestions.

Also simple _find and replace_ can fix the spacing issues `if(`

seanstrom:
@simov I would try to ignore the references to constructObject. That's really just a way for me to do batched assignment. I don't want to manually type each thing to be assigned, I rather have something wrap that up for me. That's not necessarily constructObject but at the moment constructObject is still helping with the refactor, so I plan on keeping it there until I finish up the rest of the refactor.

What would be nice is if people looking at the code can point out possible flaws in behavior. Since this code is a pretty sensitive area.

seanstrom:
quick update: I have moved around several pieces of code and restructured stuff.
Im still refactoring the callback functions for piping and nexttick stuff.

Would like some review here around the behavior, the tests at the moment but I don't think the tests are bullet proof enough to say I can trust that everything is working the way it should. Another set of eyes should take a look.

simov:
I'm glad you removed the extend/done stuff.

This is how you can handle errors without introducing additional layer of abstraction https://github.com/simov/request/commit/ccc95a8d495271e77b4d6350856b0343dfeedfda I hope it makes sense. I'll point out a few other things in the comment below, as I'm not sure you're going to pull this anyway.

Most of the variables on top of the init method (where the `// Setup Initial Values` comment is) are there just to make the code looks how you want, but it doesn't make the code easier to maintain. I'm giving you an example: `var qsLib = self.qsLib || (options.useQuerystring ? querystring : qs)` then `self.qsLib = qsLib` - just use `self.qsLib = (options.useQuerystring ? querystring : qs)` it's exactly the same! No need for indirection here.

You only need just a few separate functions, I can tell them by name if you want (don't ask me how I know, we discussed similar issues in your previous PR about refactoring). I'm giving you an example:

```js
function urlToUri(uri, url) {
  if (!uri && url) {
    uri = url
  }

  return uri
}
```

Really? How is this supposed to be reused in _any_ scenario? So I need to search for that function each time I read the code (400 lines above) just to be sure what is going on. (and that's just a single `if` statement)

As for checking the behavior - that's not the point of the refactoring, if our tests are not good enough we'll catch the error afterwards anyway. (similar thing happened when I refactored the auth logic)

Also can you rebase all of this commits into one when you feel you don't need them anymore, because they certainly doesn't make sense for anyone else.

seanstrom:
Can you point out how returning an object with properties in an function is extra abstraction?
I can see it being a different convention but I don't see how it sits at a drastically different level of abstraction.

Also part of why I didn't use a try catch block was for a optimization purpose. That's not really a good reason if we don't care about that optimization though.

Also where did anyone learn that method refactoring only applies to duplication? There's been plenty of times I've seen it used for code clarity.

In the example you brought up, urlToUri is function that will get the correct property off self. I didn't want the Init function to care about how I got that data l, it justs calls function. The reader should be able to tell by the name function what it is doing.
If not they can search for the function and find out.
If you want I can move the function closer but I'm sure that wasn't your point.

That said I am I'm partial to inlining that function being that's only a single if. But my points should support the existence of that function and others.

My point about the behavior is that we probably have brittle tests. And refactoring on brittle tests can just create a bunch of bugs. In the context of TDD, refactoring happens as the last step, after we've written tests and code.

In this context it feels like we've written some tests and mostly code, and now we want to cleanup it up. Without strong tests the cleanup phase can effectively be normal changes, since we may not know if I'm making breaking changes.
I think we should take a little more time understanding the diff but we can also just make our users tell us if we have bugs from this.

As to your comment about the variable assignments, I agree it looks messy, and I considered changing the way I'm writing it.
I'll get back to you on that one, with either the change or a different proposal.

As for the commits, I was planning on rebasing them, but how about we take a quick pole to see if they're useful or makes sense for anyone? :)

simov:
You don't really need a try/catch either, I'm just refactoring your code, just pass self. Take a look at the rest of the _modules_ that I refactored - they all use self, and as far as I can tell that was received well, because we already had a couple of contributions to these files.

Use functions to capsulate some meaningful amount of code, instead of making it look fancy.

Also feel free to close this PR, and open up another, hopefully everyone learned something from this thread, so a fresh start might be a better option instead of digging here.

Also here is how my remotes are looking atm

```bash
$ g remote -v
nylen	git@github.com:nylen/request.git (fetch)
nylen	git@github.com:nylen/request.git (push)
request	git@github.com:request/request.git (fetch)
request	git@github.com:request/request.git (push)
seanstrom	git@github.com:seanstrom/request.git (fetch)
seanstrom	git@github.com:seanstrom/request.git (push)
simov	git@github.com:simov/request.git (fetch)
simov	git@github.com:simov/request.git (push)
```

- `request` is mostly for pulling, I very rarely push there
- `simov` is for pushing only
- `nylen` `seanstrom` and anyone else is for pulling and testing
- remotes are cheap

So when I see a PR from you I expect to fetch the branch from your remote. That way you can keep record of all your branches while at the same time keeping the main repo clean. Very rarely we actually need additional branch in the main repo, and still the effort should be synchronized, so it's better to keep things separated in the first place. That doesn't apply for Mikeal ofc :)

seanstrom:
So to address your concern about this branch being incorrectly started from master.
I agree, I didn't catch this on my dev, I just blindly made a branch, Im fine with moving this somewhere else.

seanstrom:
>
You don't really need a try/catch either, I'm just refactoring your code, just pass self. Take a look at the rest of the modules that I refactored - they all use self, and as far as I can tell that was received well, because we already had a couple of contributions to these files.
>
Yeah I would believe that people would use self if they don't agree with why I don't want to mutate self in my functions, aside from `initRequest`.
Note: It may not be that your code is well received, but at the moment it's checked in, so we are more or less stuck with the way you did something because you put in the work.

>
Also feel free to close this PR, and open up another, hopefully everyone learned something from this thread, so a fresh start might be a better option instead of digging here.
>
Maybe others have learned something, perhaps its how stubborn I am. Or even how stubborn your are.
The only thing I have learned from this is that it feels your opinion is law. When to me its just opinion, and one that I disagree with. I think the fact that most of your comments are nitpicky coding conventions than commenting on the actual changes in behavior, between the initRequest function and the rest of the functions, shows how bent your are to having things your way (or some way).

The project shouldn't be locked down to a certain coding style or mindset.
I try to adhere to the current conventions, but I also try to evolve the standard.
Especially since the code itself may need another mindset/pov in order to progress well.
I personally think that it's hard for you to reason to me as a peer because you already think Im wrong.
And since your in a position of control, you can see to it that my contributions conform to your opinion.
Even though opinion here has mostly been about layers of abstraction and method extraction, which is far from actual behavior and true Architecture in the code base.

On top of that, most of the time you haven't properly responded to my questioning of your opinion and at best had straw man arguments to support yourself.

I take this quote of yours to be suggesting to throw away work, because we conflict in minor coding opinions? I would like others to chime in to if that's truly appropriate.
I wouldn't expect most of these behaviors or suggestions from this project, given its Open Governance model. In reality your actions seem to reflect something more closed off than open.

simov:
I kind of expected that response, but since you addressed it here, here is my response:

My refactoring suggestions were approved by at least two other contributors, while yours were not - not only by me, just check the comments from above.

Sean, whatever arguments I pull up, you always contra argument me. How can I suggest something without suggesting it? It's always too hard, it's always my way, and btw everyone else's opinion is wrong too, right?

You are taking everything too personally, my comments are about the project, not about you, it feels like I'm trying to attack you most of the time, which is silly.

seanstrom:
I think my response properly reflects my frustration towards the review in this thread.
You've made plenty of suggestions, some I've agreed with, and some I haven't, and some I've argued against the reasoning for the sake that the reasoning was purely preference (not fact) and should be treated as such.

It seems like you're over exaggerating my position here, I'm only disagreeing with some of the things you've said and your overall attitude and review process.
I recall one person agreeing with you, @nylen, who I also have had similar discussions with.
It's no surprise that he would agree with your opinion since it seems he shares some of it.

I don't think your actively attacking me, I think you disagree with me to the point where I possibly won't be able to merge this in unless I make all the changes you want. Which I don't think is how the review process should be. It may seem silly to feel that way, but given the way this discussion is heading that seems like an accurate assessment.

simov:
It's about simplifying things, in your case clearly it's not. If that's not obvious to you, I can't help you, experience can't be learned from words, words carry only the meaning, not what I'm putting behind them. What can be learned though is to distinguish between [attitude and competence](http://www.catb.org/esr/faqs/hacker-howto.html#believe5) but that's just some silly document right?

mikeal:
First of all, this thread/review has gotten unnecessarily hostile. This is disappointing and not productive.

Second, my apologies for not being involved earlier. This is not the first attempt to break this function out and this approach suffers from all the same problems previous attempts suffer from so I wish I could have stepped in earlier and saved everyone a lot of work.

As I've said elsewhere, moving a function in to a new file doesn't necessarily make it simpler. In fact, it can make it a lot more confusing if there isn't a natural separation between that function and the rest of what is going on in the file it was in. In this case, `init` is fundamental to the rest of the request object and moving it to another function moves a lot of information you need when reading the rest of the implementation. I'm all for breaking out certain *features* in to their own files because those features tend to stand on their own and reading their logic and state changes aren't fundamental to understanding other parts of "core request" but there is some set of functionality that is the base of request and that should live in a single file.

Furthermore, this approach makes a few additional attempts to abstract the logic but the end results is that the implementation is *longer* with the abstractions than without them which is hugely problematic.

And lastly, this is a bad time in the history of node.js/io.js to be abstracting code that can touch timers and nextTick() because there's a bunch of changes and even bugs in both 0.12 and io.js behavior in this area.

seanstrom:
So it's been a little while since a last commented on this, and I've had plenty of time to relax and think about what I was saying and what I was hearing on this thread. Overall I feel I wasn't being constructive or collaborative. I feel that I should communicated my opinions better, and received criticism better. I'm sorry for how I acted.

simov:
@seanstrom :+1: 

All - I'm closing this PR based on previous comments about it + the logic in index.js and helpers.js changed quite a bit since then :zap:

PR #1494, opened 2015-03-19
Update eslint

froatsnook:
@simov I saw you mention updating eslint and decided to have a go at it.

In general I tried to avoid rocking the boat.  There are some new options which raise some issues since the code style was not consistent.  I just disabled these options in .eslintrc for now.

[no-shadow](http://eslint.org/docs/rules/no-shadow.html)
Sounds like a good idea to me.  There are 21 no-shadow issues in the code right now.

[key-spacing](http://eslint.org/docs/rules/key-spacing.html)
At the moment the codebase uses a mixes of the two styles, which means eslint can't help us.

[no-multi-spaces](http://eslint.org/docs/rules/no-multi-spaces.html)
There are a lot of inconsistencies in the code with this one:
```
// test-baseUrl.js
var http = require('http')
  , request = require('../index')
  , tape = require('tape')
  , url = require('url')
```

vs

```
// test-agentOptions.js
var request = require('../index')
  , http    = require('http')
  , server  = require('./server')
  , tape    = require('tape')
```

as well as:

```
// lib/getProxyFromURI.js:67
    return process.env.HTTPS_PROXY ||
           process.env.https_proxy ||
           process.env.HTTP_PROXY  ||
           process.env.http_proxy  || null
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,8 @@
     "node": true
   },
   "rules": {
+    // 2-space indentation
+    "indent": [2, 2],
     // Disallow semi-colons, unless needed to disambiguate statement
     "semi": [2, "never"],
     // Require strings to use single quotes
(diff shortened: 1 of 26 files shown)


simov:
Thanks, @froatsnook that might come in handy. I like that approach of introducing it with minimum changes to the code. Also only 3 additional disabled options is not that bad I think. What I wanted to add was `"indent": [2, 2]` for the 2 space indentation, because it is disabled by default.

froatsnook:
@simov Good call.  That found a few indentation errors.

FredKSchott:
+10000 So glad we can finally update out of node v0.8 purgatory

simov:
This PR is looking good to me, we can think about enabling some of the `eslint` options later. The only thing that I'll suggest is to rebase all commits into one, as they are too small to have any meaning on its own.

froatsnook:
@simov OK, done!

simov:
Thanks @froatsnook :+1:

PR #1496, opened 2015-03-20
Parameters encoded to base 64 should be decoded as UTF-8, not ASCII.

albanm:
This fixes a bug that I have currently. Some users were created with special characters in password but could not authenticate through request.auth method afterward.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -74,7 +74,7 @@ function isReadStream (rs) {
 }
 
 function toBase64 (str) {
-  return (new Buffer(str || '', 'ascii')).toString('base64')
+  return (new Buffer(str || '', 'utf8')).toString('base64')
 }
 
 exports.isFunction            = isFunction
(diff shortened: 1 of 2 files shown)


froatsnook:
Is UTF-8 the right encoding to use, though?  Or should it be ISO-8859-1?  Apparently there's no consistency among the spec, browser implementations, and server implementations: https://code.google.com/p/chromium/issues/detail?id=25790

I would say we should use utf-8 as you suggested, but also accept a `Buffer` for `user` and/or `pass` for users dealing with servers that expect ISO-8859-1 instead of utf-8.

simov:
Thanks @albanm ! Also the remarks about ISO-8859-1 are correct, but I think we should be good with just `utf8` for now :+1:

albanm:
Thanks !

Is it possible to release a small patch version containing this fix ?

simov:
The next request release is scheduled for the upcoming week, I can't tell you exact day and time, but we're rounding up the last features/bugfixes and then we're good to go.

Issue #1340, opened 2015-01-06
Restructure documentation


nylen:
Our README has gotten really long and I think it's time to split it into separate files (see discussion at #1331).  Here are the sections in it now:

```sh
$ grep '^## ' README.md
## Super simple to use
## Streaming
## Proxies
## UNIX Socket
## Forms
## HTTP Authentication
## OAuth Signing
## Custom HTTP Headers
## TLS/SSL Protocol
## request(options, callback)
## Convenience methods
## Examples:
## Debugging
```

Here's how I would reorder and modify these sections:

```sh
## Super simple to use
add link to full list of options
## Streaming
short explanation with link to `doc/streaming.md`
## Forms
short-to-medium explanation with link to 'doc/forms.md'
## HTTP Authentication
short explanation with link to 'doc/auth.md'
## Custom HTTP Headers
this one is fine
## OAuth Signing
short explanation with link to 'doc/oauth.md'
## Proxies
short explanation with link to 'doc/proxies.md'
## UNIX Socket
change name to "Unix Domain Sockets"
## TLS/SSL Protocol
short explanation with link to 'doc/ssl.md'
## request(options, callback)
change name to "All Available Options" or similar
we could also move the full API to its own file, and
give each option its own section?  this would allow
linking directly to individual options
## Convenience methods
this should go wherever the full API docs go
## Examples:
move these to the 'examples/' dir
and have a readme describing the examples,
with each one in its own .js file
## Debugging
this one is fine
```

These are just some first thoughts, so feedback welcome.  @rockbot I think this helps address your concern by making it easier/quicker to find the full options we support, but I want to keep the narrative of "here's what this library can do... OK finally here's how you do all these things and more."

simov:
:+1: for splitting the docs into separate files, I also have a module that turns [this](https://github.com/simov/purest/tree/gh-pages/markdown) into [this](http://simov.github.io/purest/html/) and it have single page version of that, but anyway that's a very obsolete verison of it and it's still not published. There might be other modules for that too.

FredKSchott:
+@mikeal

A full top-to-bottom README restructure/rewrite and documentation clean up could be a great task for that bounty mentioned in #1122 
To get someone with some serious doc-writing experience and a well-trained grasp on how to use and explain request to both new and expert users.

rockbot:
@nylen I'm wary about splitting the options out into a separate file, mostly because then from a UX perspective you're forcing the user to leave and then have to remember where they were before they got there.

[...]

PR #1498, opened 2015-03-21
Add table of contents in readme

simov:
- Add table of contents in readme
- Add back to top links at the end of each section
- Add horizontal line divider after each section
- Add visual cues in options list

Closes https://github.com/request/request/issues/1340
Also related to https://github.com/request/request/pull/1331

Basically I refactored the readme according to what @nylen outlined in https://github.com/request/request/issues/1340#issuecomment-69367275 as collaborative effort to improve the readme.

Easier to review the this PR would be to just check out the rendered markdown file https://github.com/simov/request/tree/refactor-readme

Table of contents and back to top links at the end of each section are pretty self explanatory. The horizontal line divider between the sections is mostly for easier maintenance of the file (if you have good markdown syntax highlighting you'll now what I mean) Also I'm putting 2 blank lines before and after each horizontal rule, and two blank lines before each header (not an actual rule, just fyi)

The last thing was to use simple horizontal lines again, this time for the list of options.

Now if you take a look at it, you'll notice that it's a bit easier to fix your sight on certain things. So the presentation remains linear (no additional headers or text) yet just enough to scan through it faster.

I also wanted to split the sections into separate files and put them under docs folder. But I couldn't figure out a good way to introduce the building process to the user while having the combined readme all the time too.

/cc @rockbot @nylen @FredKSchott and anyone else

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,18 @@
-# Request — Simplified HTTP client
+
+# Request - Simplified HTTP client
+
 [![npm package](https://nodei.co/npm/request.png?downloads=true&downloadRank=true&stars=true)](https://nodei.co/npm/request/)
 
-[![Build status](https://img.shields.io/travis/request/request.svg?style=flat)](https://travis-ci.org/request/request)
-[![Coverage](https://img.shields.io/coveralls/request/request.svg?style=flat)](https://coveralls.io/r/request/request)
-[![Gitter](https://img.shields.io/badge/gitter-join_chat-blue.svg?style=flat)](https://gitter.im/request/request?utm_source=badge)
+[![Build status](https://img.shields.io/travis/request/request.svg?style=flat-square)](https://travis-ci.org/request/request)
+[![Coverage](https://img.shields.io/coveralls/request/request.svg?style=flat-square)](https://coveralls.io/r/request/request)
+[![Gitter](https://img.shields.io/badge/gitter-join_chat-blue.svg?style=flat-square)](https://gitter.im/request/request?utm_source=badge)
+
 
 ## Super simple to use
 
 Request is designed to be the simplest way possible to make http calls. It supports HTTPS and follows redirects by default.
 
-```javascript
+```js
 var request = require('request');
 request('http://www.google.com', function (error, response, body) {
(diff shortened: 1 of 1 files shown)


FredKSchott:
+1 I like it

nylen:
**This is awesome**.

dance

I'm merging so it doesn't go stale.  A few points of feedback, though:

## Horizontal lines

I'm not super excited about all the `<hr>`s.  I think these should just be headings instead, since GitHub already provides separation:

screen shot 2015-03-21 at 6 04 49 pm
screen shot 2015-03-21 at 6 05 02 pm

The splits in the all-options section should be named headers too, but at a lower level, like `####`:

screen shot 2015-03-21 at 6 05 37 pm

## Linking to individual options

There are lots of places in the docs where we cross-reference options, and making them clickable links would be a nice enhancement.  Last time I checked (Jan 2015) the following markup works on GitHub:

```html
<a href="#user-content-something" id="something">some text</a>
<a href="#user-content-something" name="something">some text</a>
```

Examples:
- https://github.com/nylen/playground/blob/patch-1/anchors.md#user-content-dsazf
- https://github.com/nylen/playground/blob/patch-1/anchors.md#user-content-sdaf

## Huge images

Images in this post are huge because retina display, 
deal-with-it

nylen:
Forgot to say - I think using headers & spacing addresses the maintainability concerns too, since good Markdown syntax highlighting applies to headers.

simov:
I intentionally didn't want to add additional headers to the options list, but yeah if someone show me a good example of using them, I can change my mind eventually. Btw H6 is a grayed out header according the the GitHub's css:

###### Wow I'm a header too

As for the horizontal rules separating the sections - I can live without them, if that's such an issue.

Huge images - nice! It's like me wearing glasses :trollface:

nylen:
Not a big deal, it's still a huge improvement.  Just my $0.02:  it's graphically cleaner to use only headers rather than both, and semantically clearer to use named sections rather than lines.  Named sections also makes the best place for new options more clear.

nicolasmccurdy:
@nylen :+1:

PR #1500, opened 2015-03-22
Adding handling for no auth method and null bearer

philberg:
I noticed a couple of scenarios the other day:

0.  When no authentication mechanism is defined, or bearer = undefined (I meant to send it, but it got set incorrectly) the error message is difficult to deciper
0.  When setting `bearer: null`, request does a string concat, which results in `'bearer null'`. Probably not the intended behavior

Also added explanations to the bearer tests

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -42,7 +42,7 @@ Auth.prototype.bearer = function (bearer, sendImmediately) {
     if (typeof bearer === 'function') {
       bearer = bearer()
     }
-    var authHeader = 'Bearer ' + bearer
+    var authHeader = 'Bearer ' + (bearer || '')
     self.sentAuth = true
     return authHeader
   }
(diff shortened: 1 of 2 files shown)


philberg:
That line will evaluate to `bearer ` when the bearer is set to `null` (but no undefined, because that enters the basic auth case).

My intent was to make the behavior less confusing.
Currently, setting `bearer: null` actually sends the string null as the bearer.  My interpretation was that, by setting `bearer: null` the intent was to send bearer with nothing, in other words `bearer `

If the intent is so actually send the string null, then `bearer: 'null'` can be set

simov:
Ok, what I'm saying is that you can remove the `req.headers.authorization === 'Bearer null'` check from the test because it's misleading and it's never true.

Other than that this PR looks good to me :+1:

philberg:
I operate under the principle that a change should never be made without a corresponding test, so I was trying to write a test that would have failed before but now will pass.  I'm sure there is a more elegant way to do it.

Either way, I understand what you're saying, so I've removed that line and rebased into the same commit

nylen:
> I operate under the principle that a change should never be made without a corresponding test, so I was trying to write a test that would have failed before but now will pass.

+1000 :bow:

In this case, your [`t.throws` test](https://github.com/request/request/pull/1500/files#diff-19cb0f54b8954f938e0d28dee1b90b37R151) would have failed before, so I think you're good.

nylen:
Thanks for adding test names, too.

Issue #1506, opened 2015-03-25
undefined is not a function (functions must be bound by 'request').


noseglid:
This worked in v2.53.0

```javascript
var Promise = require('bluebird');
var postAsync = Promise.promisify(require('request').post);

postAsync('http://www.google.com').then(function (content) {
  console.log(content);
});
```

in v2.54.0, above code gives this error:

```
Unhandled rejection TypeError: undefined is not a function
    at request.(anonymous function) (/home/alexo/tmp/node_modules/request/index.js:64:16)
    at tryCatcher (/home/alexo/tmp/node_modules/bluebird/js/main/util.js:24:31)
    at ret (eval at <anonymous> (/home/alexo/tmp/node_modules/bluebird/js/main/promisify.js:154:12), <anonymous>:13:23)
    at Object.<anonymous> (/home/alexo/tmp/a.js:4:1)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
```

Any ideas what changes might've caused this?

nicolasmccurdy:
I assume your file is `/home/alexo/tmp/a.js`?

noseglid:
Investigated this a bit more. It seems that request.[verb] now requires it to be bound by 'request'.

This example doesn't work either:

```javascript
var post = require('request').post;
post('http://www.google.com', function (err, response, body) {
  console.log(body);
});
```

output:
```javascript

/home/alexo/tmp/node_modules/request/index.js:65
    return this(params.uri || null, params.options, params.callback)
               ^
TypeError: undefined is not a function
    at request.(anonymous function) (/home/alexo/tmp/node_modules/request/index.js:65:16)
    at Object.<anonymous> (/home/alexo/tmp/a.js:5:1)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
```


The promisifying can be solved by doing: `var postAsync = Promise.promisify(require('request').post.bind(request))` . Then it works.

ericlu88:
+1 for experiencing the same problem. The change seems to be introduced in #1430

[...]

PR #1515, opened 2015-03-27
Allow static invoking of convenience methods

simov:
Fixes https://github.com/request/request/issues/1506

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,7 @@ verbs.forEach(function(verb){
   request[verb] = function(uri, options, callback){
     var params = initParams(uri, options, callback)
     params.options.method = method
-    return this(params.uri || null, params.options, params.callback)
+    return (this || request)(params.uri || null, params.options, params.callback)
   }
 })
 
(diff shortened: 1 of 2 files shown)


kytwb:
Can we get a release for this asap?

simov:
@kytwb we're rounding up a few more bugfixes and improvements, so until then you can use version 2.53 unless your code relies on features introduced in 2.54, in that case you can use the latest master in your package.json (not perfect)

```js
"request": "git://github.com/request/request.git"
```

kytwb:
@simov: Most of our projects build is now broken due to this regression, would rather wait for a release if it's coming soon than changing the package.json in all the projects to finally rollback when the fix is out. Let me know.

simov:
Ok, my assumption was that you have fixed dependency versions in your packages, which again my not be ideal but it's safer.

Anyway I understand that this bug is a problem for a lot of people, so it's better to have a new release soon. However there is one more bugfix submitted related to the same issue, then I'm going to make some improvements on that logic that caused the bugs. So we should wait a couple of days, probably till the end of the week.

kytwb:
@simov: Fair enough, thanks for the prompt reply.

simov:
2.55 is published on NPM

PR #1520, opened 2015-03-30
Refactor defaults

simov:
This is a follow up to https://github.com/request/request/pull/1518

Much more simplified defaults and params initialization in index.js. Basically flattened the logic, also moved the `wrap` closure outside of the `defaults` method and I left it in index.js because I think it's very closely related to the `request` method, and the convenience http verb methods. Also completely dropped all of the object extend closures, as I don't think they add any substantial value to that piece of code.

/cc @nylen @FredKSchott @mikeal

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,22 +19,30 @@ var extend                = require('util')._extend
   , helpers               = require('./lib/helpers')
 
 var isFunction            = helpers.isFunction
-  , constructObject       = helpers.constructObject
-  , filterForCallback     = helpers.filterForCallback
-  , constructOptionsFrom  = helpers.constructOptionsFrom
   , paramsHaveRequestBody = helpers.paramsHaveRequestBody
 
 
 // organize params for patch, post, put, head, del
 function initParams(uri, options, callback) {
-  callback = filterForCallback([options, callback])
-  options = constructOptionsFrom(uri, options)
+  if (typeof options === 'function') {
+    callback = options
+  }
 
-  return constructObject()
-    .extend({callback: callback})
-    .extend({options: options})
-    .extend({uri: options.uri})
-    .done()
+  var params
+  if (typeof options === 'object') {
+    params = extend({}, options)
+    params = extend(params, {uri: uri})
+  } else if (typeof uri === 'string') {
+    params = extend({}, {uri: uri})
+  } else {
+    params = extend({}, uri)
+  }
+
+  return {
+    uri: params.uri,
+    options: params,
+    callback: callback
+  }
 }
 
 function request (uri, options, callback) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -73,77 +81,65 @@ request.cookie = function (str) {
   return cookies.parse(str)
 }
 
-request.defaults = function (options, requester) {
+function wrap (method, options, requester) {
```
FredKSchott: `wrap` is a little vague, it took me a while to understand what this function was doing. Is there a better name here that could provide a little more context? `wrapDefaults`? `wrapRequestMethods`?

Also, some documentation on expected arguments would be equally helpful, especially since it looks like we support a lot of different ways to call this (method = string, function, or request, request is optional, etc).

index.js
```diff
@@ -73,77 +81,65 @@ request.cookie = function (str) {
...
-        method = requester
-      }
+  var defaults      = wrap(self, options, requester)
 
-      return method(params.options, params.callback)
-    }
-  }
+  var verbs = ['get', 'head', 'post', 'put', 'patch', 'del']
+  verbs.forEach(function(verb) {
+    defaults[verb]  = wrap(verb, options, requester)
```
FredKSchott: Could we change how we call this function to simplify wrap's behavior? The complexity of supporting both functions and strings for the first argument would be unnecessary if you could always pass the function you want to wrap.

lib/helpers.js
```diff
@@ -12,40 +11,10 @@ function deferMethod() {
   return setImmediate
 }
 
-function constructObject(initialObject) {
```
FredKSchott: Nice! +1 on removing this custom implementation and using normal javascript objects

index.js
simov: Yep I agree, crossed my mind, but I decided to keep it that way, though I haven't tested how the code will look like with it. So I'll give it a try.

index.js
```diff
@@ -73,77 +81,65 @@ request.cookie = function (str) {
...
-    options.extend(optionsArg)
+    options = extend({}, optionsArg)
   }
   if (agentOptions) {
     options.agentOptions = agentOptions
   }
 
-  options.extend({forever: true})
-  return request.defaults(options.done())
+  options = extend(options, {forever: true})
```
FredKSchott: could this just be written as `options.forever = true`?

FredKSchott:
:+1: with comments

simov:
I managed to squeeze the code a bit further by removing redundant data structure returned by `initParams`

Regarding your comment @FredKSchott about passing a string to the `wrapRequestMehod` function, I can't figure out a good way to remove the string parameter from there without introducing complexity in another place.

The reason for passing string there is to be able to set the `params.method` which is later used in the `request` function to abort the request on `HEAD` + body.

Similar to the way the http verb methods set the `method` name after calling `initParams`, the `wrapRequestMethod` is doing exactly the same, again after the `initParams` was called.

Here is the catch - on wrapped http verb method + requester, the default http verb method is never called, because the requester takes precedence. Therefore the request `method` is never set, so the request can't be aborted in the `request` function (and we have a test for that too).

Issue #1509, opened 2015-03-26
options.uri is a required argument when using default uri


neverfox:
I make heavy use of the `defaults` method to to create a request object with a fixed URI and then setting the other options dynamically. As of 2.54.0, I'm now getting `options.uri is a required argument` when trying to make a request on that object (where I've already set the uri). Downgrading to 2.53.0 clears it up. I didn't see anything in the changelog that would indicate that this would break.

simov:
It was introduced here https://github.com/request/request/pull/1469 @froatsnook can you take a look at it? :)

froatsnook:
Oops! I'm on it.

froatsnook:
@neverfox I can't reproduce this.

```
$ npm show request | grep 'version:'
  version: '2.54.0',
$ node request-test-defaults.js
OK
OK
```

Could you please run the test program below and let me know whether prints OK or FAIL?  If it prints OK, could you modify it to reproduce your error?  Or provide some other example that breaks?

```
'use strict'

var http = require('http')
  , request = require('request')
  , url = require('url')

var s = http.createServer(function(req, res) {
  res.statusCode = 200
  res.writeHead(200, { 'Content-Type': 'application/json' })
  res.end(JSON.stringify(req.headers))
})

s.listen(6767)

var r = request.defaults({
  uri: 'http://localhost:6767/test'
})

var count = 0

r({
  json: true,
  headers: {
    'X-TEST': 'test 1'
  }
}, function(err, res, body) {
  if (err || body['x-test'] !== 'test 1') {
    console.log('FAIL')
  } else {
    console.log('OK')
  }

  if (++count === 2) {
    s.close()
  }
})

r({
  json: true,
  headers: {
    'X-TEST': 'test 2'
  }
}, function(err, res, body) {
  if (err || body['x-test'] !== 'test 2') {
    console.log('FAIL')
  } else {
    console.log('OK')
  }

  if (++count === 2) {
    s.close()
  }
})
```

neverfox:
Yes, I will. Thanks for the code.

simov:
Actually @neverfox your bug report is a bit vague:

> I make heavy use of the defaults method to to create a request object with a fixed URI and then setting the other options dynamically.

Can you give us a short code example of what are you using, and what you think might cause the issue?

neverfox:
Here is the test code above, modified with some code from my project that recreates the error:

```javascript
'use strict'

var Promise = require('bluebird')
  , http = require('http')
[...]

PR #1521, opened 2015-03-31
Add promise tests

simov:
A few promise tests using [bluebird](https://github.com/petkaantonov/bluebird) as it seems to be the most popular promise solution atm, plus it's used by [request-promise](https://github.com/tyabonil/request-promise) internally.

This is in attempt to make the promise support first class citizen in request, as it turns out that a lot of people are using promises with request. Also these tests should cover the bugs found in https://github.com/request/request/issues/1506 and https://github.com/request/request/issues/1509

@kytwb @noseglid @ericlu88 @carsy @neverfox let me know if there is anything else that can be added as test cases here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "rimraf": "~2.2.8",
     "server-destroy": "~1.0.0",
     "tape": "~3.0.0",
-    "taper": "~0.4.0"
+    "taper": "~0.4.0",
+    "bluebird": "~2.9.21"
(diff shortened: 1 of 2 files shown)


noseglid:
It looks like good coverage. It covers the use cases we have at least. I'm glad to see this addressed! :+1:

simov:
Thanks for the feedback @noseglid :+1:

PR #1533, opened 2015-04-09
Adding dependency status bar to README.md

YasharF:
Adding dependency status bar to README.md to make it easier to identify when one or more dependencies are out of date.


simov:
That might come in handy, but you should use icon from shields.io instead

YasharF:
Oh, I see: https://img.shields.io/david/request/request.svg?style=flat-square , I'll fix it up tonight or tomorrow.

YasharF:
Updated to the svg from shields.io

PR #1536, opened 2015-04-09
More eslint

froatsnook:
I updated eslint to 0.18.0 and added a couple rules for style consistency.

```
    // Use if () { }
    //       ^ space
    "space-after-keywords": [2, "always"],
```

There were only a few instances of `if()` and `else{` in the code, so this was clear.

```
    // Use if () { }
    //          ^ space
    "space-before-blocks": [2, "always"],
```

There were only a few instances of `if (){` in the code, so this was also clear.

```
    // Use function test() { }
    //                 ^^ no space
    "space-before-function-paren": [2, "never"]
```

This one was harder.  There were tons of `function ()`s and `function name ()`s in the code, including `function Request ()`.  But there were way more `function()`s and `function name()`s.  Introducing with `"always"` created 739 errors, vs `"never"` creating 245.

One additional change (1ffc50d53511eb1b81d64da816adad3aa687e6b0): eslint complained about an empty block statement, so I added the comment `// empty` to explicitly mark it as empty.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -34,6 +34,9 @@
     // eslint can't handle this, so the check is disabled.
     "key-spacing": 0,
     // Allow shadowing vars in outer scope (needs discussion)
-    "no-shadow": 0
+    "no-shadow": 0,
+    // Use if () { }
+    //       ^ space
+    "space-after-keywords": [2, "always"]
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -37,6 +37,9 @@
     "no-shadow": 0,
     // Use if () { }
     //       ^ space
-    "space-after-keywords": [2, "always"]
+    "space-after-keywords": [2, "always"],
+    // Use if () { }
+    //          ^ space
+    "space-before-blocks": [2, "always"]
   }
 }
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -4,7 +4,7 @@ var jsonSafeStringify = require('json-stringify-safe')
   , crypto = require('crypto')
 
 function deferMethod() {
-  if(typeof setImmediate === 'undefined') {
+  if (typeof setImmediate === 'undefined') {
     return process.nextTick
   }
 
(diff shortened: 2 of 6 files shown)


simov:
That's great, I'm only not too sure about the `space-before-function-paren` rule.

If you run (without the tests)
```
node node_modules/.bin/eslint lib/ *.js
```
with `"space-before-function-paren": [2, "never"]` you get 99 warnings

with `"space-before-function-paren": [2, "aways"]` you get 41 warnings

So it turns out that there are more occurrences are using space, of course without the tests, but they were largely refactored by one person at some point.

So I'm for keeping the things as they are for now regarding this rule.

froatsnook:
> So I'm for keeping the things as they are for now regarding this rule.

Do you mean set it to always?  Or leave the rule off?

simov:
I was thinking about leaving it off, which is not perfect either, and it will be a matter of personal preference whichever we choose.

froatsnook:
I would say we should just pick something.  This is a much less polarizing style issue than some of the other rules.

simov:
I agree that we should pick one, but which one exactly? For example I'm using spaces always, so it would be perfectly fine for me to have all places changed to using spaces. Just by looking at the diff I can say that other developers have the same preference too.

On the other hand saying that you should always use space might not be appealing to everyone, so that's why my suggestion was to leave that rule off for now.

But lets try to make a quick pool.

@mikeal @FredKSchott @nylen pick either 1) or 2)

1)

```js
function()
function func(param)
```

2)

```js
function ()
function func (param)
```

simov:
Ok, so I'm sticking to my initial decision about this one to keep the `space-before-function-paren` as it was up until now. I don't mind any of the two spacing schemes, but I'm not ok with forcing everyone to use either one.

So, @froatsnook can you revert only this https://github.com/froatsnook/request/commit/22870c7028c0f4d67239c5437f1384cf6f777a2f commit?

froatsnook:
Sure, will do.

froatsnook:
OK, reverted.

simov:
Thanks :+1:

PR #1539, opened 2015-04-11
ensure the latest version of har-validator is included

ahmadnassri:
seems like some folks have had issues when using request from source with `har-validator` dependency (see: https://github.com/ahmadnassri/har-validator/issues/6) `v1.6.1` of `har-validator` addresses these problems.


simov:
:+1:

Issue #1318, opened 2014-12-15
Error when URI scheme is not specified


vbauer:
POST request to "http://sometime.com" sends fine, but it fails on "sometime.com":

```
/home/vbauer/workspace/js/manet/node_modules/request/request.js:914
  self.req = self.httpModule.request(reqOptions)
                             ^
TypeError: Cannot call method 'request' of undefined
    at Request.start (/home/vbauer/workspace/js/manet/node_modules/request/request.js:914:30)
    at Request.write (/home/vbauer/workspace/js/manet/node_modules/request/request.js:1722:10)
    at write (_stream_readable.js:583:24)
    at flow (_stream_readable.js:592:7)
    at ReadStream.pipeOnReadable (_stream_readable.js:624:5)
    at ReadStream.EventEmitter.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:408:10)
    at emitReadable (_stream_readable.js:404:5)
    at readableAddChunk (_stream_readable.js:165:9)
    at ReadStream.Readable.push (_stream_readable.js:127:10)
```

Maybe it will be better to use "http" as default scheme?

saper:
Is this still an issue? I am getting this:

````
> var r=require('request');
undefined
> var p=r.get('sometime.com');
Error: Invalid URI "sometime.com"
    at Request.init (/home/saper/node_modules/request/request.js:469:31)
    at new Request (/home/saper/node_modules/request/request.js:272:8)
    at request (/home/saper/node_modules/request/index.js:56:10)
    at Function.request.(anonymous function) (/home/saper/node_modules/request/index.js:66:12)
    at repl:1:9
    at REPLServer.defaultEval (repl.js:132:27)
    at bound (domain.js:254:14)
    at REPLServer.runBound [as eval] (domain.js:267:12)
    at REPLServer.<anonymous> (repl.js:279:12)
    at REPLServer.emit (events.js:107:17)
> undefined
> var p=r.post('sometime.com');
Error: Invalid URI "sometime.com"
    at Request.init (/home/saper/node_modules/request/request.js:469:31)
    at new Request (/home/saper/node_modules/request/request.js:272:8)
    at request (/home/saper/node_modules/request/index.js:56:10)
    at Function.request.(anonymous function) (/home/saper/node_modules/request/index.js:66:12)
    at repl:1:9
    at REPLServer.defaultEval (repl.js:132:27)
    at bound (domain.js:254:14)
[...]

PR #1558, opened 2015-05-01
Emit error instead of throw

simov:
Related to this https://github.com/request/request/issues/1318#issuecomment-97374120 onward 

### Breaking change

Any code that previously relied on catching the errors in the `catch` block will now receive them in the `callback` as errors if a `callback` is specified

```js
request({uri:'...'}, function (err, res, body) {
  if (err) {
    // all errors are received here
  }
}).on('error', function (err) {
  // that's being called before being registered
})
```

### No callback during initialization

This repository still have two error cases that throws in index.js But if the callback is missing, any error will be thrown anyway

```js
try {
  request({uri:'...'}).on('error', function (err) {
    // that's being called before being registered
  })
} catch (err) {
  // EventEmitter throws the error if the event is unhandled
}
```

### After successful initialization

```js
request({uri:'...'}, function (err, res, body) {
  if (err) {
    // all errors are received here
  }
}).on('error', function (err) {
  // after successful initialization this is being registered
})
```

---

One possible way to receive the errors during initialization in the `on('error')` event handler, is to use this hack

```js
setTimeout(self.emit.bind(self, 'error', new Error('...')), 0)
// instead of
self.emit('error', new Error('...'))
```
But it's unreliable and I'm not going to use it.

// cc @mikeal @nylen @FredKSchott @saper

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -21,7 +21,7 @@ function Auth (request) {
 Auth.prototype.basic = function (user, pass, sendImmediately) {
   var self = this
   if (typeof user !== 'string' || (pass !== undefined && typeof pass !== 'string')) {
-    throw new Error('auth() received invalid user or password')
+    self.request.emit('error', new Error('auth() received invalid user or password'))
   }
   self.user = user
   self.pass = pass
@@ -115,7 +115,7 @@ Auth.prototype.onRequest = function (user, pass, sendImmediately, bearer) {
 
   var authHeader
   if (bearer === undefined && user === undefined) {
-    throw new Error('no auth mechanism defined')
+    self.request.emit('error', new Error('no auth mechanism defined'))
   } else if (bearer !== undefined) {
     authHeader = self.bearer(bearer, sendImmediately)
   } else {
--- a/lib/multipart.js
+++ b/lib/multipart.js
@@ -18,7 +18,7 @@ Multipart.prototype.isChunked = function (options) {
     , parts = options.data || options
 
   if (!parts.forEach) {
-    throw new Error('Argument error, options.multipart.')
+    self.request.emit('error', new Error('Argument error, options.multipart.'))
   }
 
   if (options.chunked !== undefined) {
@@ -32,7 +32,7 @@ Multipart.prototype.isChunked = function (options) {
   if (!chunked) {
     parts.forEach(function (part) {
       if (typeof part.body === 'undefined') {
-        throw new Error('Body attribute missing in multipart.')
+        self.request.emit('error', new Error('Body attribute missing in multipart.'))
       }
       if (isstream(part.body)) {
         chunked = true
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -60,7 +60,8 @@ OAuth.prototype.buildParams = function (_oauth, uri, method, query, form, qsLib)
 
 OAuth.prototype.buildBodyHash = function(_oauth, body) {
   if (['HMAC-SHA1', 'RSA-SHA1'].indexOf(_oauth.signature_method || 'HMAC-SHA1') < 0) {
-    throw new Error('oauth: ' + _oauth.signature_method + ' signature_method not supported with body_hash signing.')
+    this.request.emit('error', new Error('oauth: ' + _oauth.signature_method +
+      ' signature_method not supported with body_hash signing.'))
   }
 
   var shasum = crypto.createHash('sha1')
@@ -89,13 +90,12 @@ OAuth.prototype.concatParams = function (oa, sep, wrap) {
 
 OAuth.prototype.onRequest = function (_oauth) {
   var self = this
-    , request = self.request
 
-  var uri = request.uri || {}
-    , method = request.method || ''
-    , headers = caseless(request.headers)
-    , body = request.body || ''
-    , qsLib = request.qsLib || qs
+  var uri = self.request.uri || {}
+    , method = self.request.method || ''
+    , headers = caseless(self.request.headers)
+    , body = self.request.body || ''
+    , qsLib = self.request.qsLib || qs
 
   var form
     , query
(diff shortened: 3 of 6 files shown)


nylen:
I actually like the idea of emitting all errors to the `error` event handler for consistency, and you could use `process.nextTick` instead of `setTimeout`.  It's not ideal, but would triggering the error async actually cause any issues?  @simov what did you mean by "it's unreliable"?

simov:
@nylen `setTimeout` is unreliable because it doesn't guarantee when exactly it will be executed, although in most cases it should work just fine (it's 4ms at minimum). Now that I'm reading about it `nextTick` seems to have better performance because it doesn't involve a timer.

In our code base we have this [wrapper](https://github.com/request/request/blob/v2.55.0/lib/helpers.js#L6-L12) used only [at the end](https://github.com/request/request/blob/v2.55.0/request.js#L697) of the `init` method.

So we can have this code but it looks ugly, and in some cases it might cause issues I think.

```js
defer(self.emit.bind(self, 'error', new Error('...')))
// instead of
self.emit('error', new Error('...'))
```

ziggythehamster:
When I do error events before they can be listened in my code (i.e., constructors), I use `process.nextTick` to emit the error so that I have time to listen for the error.

Either all events/errors are emitted asynchronously, or they're thrown synchronously. Do not mix synchronous and asynchronous or [you will release Zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony).

If you do decide to throw/emit synchronously (where you can't `.on('error')`), can you please give us a flag to set that would allow us to opt-in to asynchronous errors?

FredKSchott:
I see the value in both behaviors, but I prefer what we have currently to this. The thinking is that errors that occur on initialization are caused by bad input / are programmer errors and should get more exposure, while async errors related to redirects and responses are expected errors (something can always go wrong).

saper:
I was trying to do something like that (naively):

````
var r = require('request')

r.on('error', function(er, {console.error('got ya!')}).get('badurl.com')
````
And get 'got ya' called on invalid URL for example.

You can export a module which behaves like `EventEmitter`. But `EventEmitter` constructor makes sure that event listener table in a new object instance. 

If you are okay with one bit ugly copy operations of the `_events` array, you can try this solution:

https://github.com/simov/request/pull/2

and it can be used as follows:

````
request.on('init', function(er) {console.log('starting')})
request.on('initialized', function(er) {console.log('ready to go')})
request.on('error', function(er) {
    console.log('first error handler')
})
var z = request.on('error', function(er) { 
    console.log('second error handler')
}).get('aaa');
````

which produces:

````
starting
first error handler
second error handler
ready to go
````

https://gist.github.com/saper/67d2bd6a6cd81ee9773a

I think this is a pretty crazy idea but may be we can achieve something useful this way.

nylen:
> I see the value in both behaviors, but I prefer what we have currently to this. The thinking is that errors that occur on initialization are caused by bad input / are programmer errors and should get more exposure, while async errors related to redirects and responses are expected errors (something can always go wrong).

You could work around this by emitting any initialization errors before adding the default `.on('error')` handler for the callback.  That way they have to be handled or there's an `uncaughtError`.

> crazy idea with module-level `.on('error')`

This feels a little too "magic" to me.

My other concern with making changes here - are we creating situations where we return a broken, half-initialized object?

saper:
On Mon, 4 May 2015, James Nylen wrote:

> > I see the value in both behaviors, but I prefer what we have currently to this. The thinking is that errors that occur on initialization are caused by bad input / are programmer errors and should get more exposure, while async errors related to redirects and responses are expected errors (something can always go wrong).
> 
> You could work around this by emitting any initialization errors before adding the default `.on('error')` handler for the callback.  That way they have to be handled or there's an `uncaughtError`.
> 
> > crazy idea with module-level `.on('error')`
> 
> This feels a little too "magic" to me.

Attaching a handler to an initialized Request obviously does not work... Given that we often face situation
where a whole URL is a configuration parameter, environment variable or worse user input I'd rather
have a general catch-all wrapper for initialization errors.

> My other concern with making changes here - are we creating situations where we return a broken, half-initialized object?

I think yes, but that's what we do now anyway:

- on module level a working `EventEmitter` interface is returned by `require('request')`. So no half-broken object here.

- on request level the situation stays the same as it was (callbacks, instance handlers get called)

- if no callbacks/handlers are provided `emit()` will raise an exception as it does today, so "new"
  fails.

- if there is a callback and `Request.init()` fails via `self.emit()` the error will be passed
  on to the callback and a partially usable object can be returned, but only after attempting
  the request.

This is the situation we have right now I think with the case:

    var z = request.get('aaa', function(e) {
     console.log('error');
    });
    console.log('Half-broken object', z.uri)

where we get with the current master:

    error
    Half-broken object { protocol: null,
      slashes: null,
      auth: null,
      host: null,
      port: null,
      hostname: null,
      hash: null,
      search: null,
      query: null,
      pathname: 'aaa',
      path: 'aaa',
      href: 'aaa' }

We never return an object straight from the constructor
before making the request, do we?

simov:
@saper your solution is clever, but modifying the global state is extremely bad practice. The problem is that you're not using the `new` keyword. I'm giving you a short [example](https://gist.github.com/simov/eff7e508bd16ee5eeab4) (run app.js)

saper:
Of course this is expected. We are simply doing

     var r = require('request').on('error', function /*.. */)

There is no way such a handler would magically be bound to the `Request` instance. It's a per-module handler, so it's neither `file1` nor `file2` nor `callx` based.

`request()` is a factory function itself.

I just had a look at node's [http module](https://nodejs.org/api/http.html) and they solve it nicely: `_http_client.ClientConnect` is synchronous and throws exceptions, while `_http_agent.Agent` is event-driven.

simov:
It is expected for single error event to pop up on various places through the entire app, and multiple times - that's some weird expectations TBH :)

saper:
I have to agree :)

well, I never really expected we will go this way - it was just a funny exercise in JavaScript.

But I agree `nextTick` is a horrible solution and it's better to say with exceptions.

simov:
Ok, so this tiny PR makes the error behavior consistent only when you have a callback. I'm glad we raised the awareness about the error related issue, but like I said a few comments above, I'm not sure about the best way to wrap the emit statements, so if there aren't any other suggestions I'm going to merge this PR shortly.

PR #1560, opened 2015-05-05
Update combined-stream

apechimp:
This updated combined-stream now uses an updated delayed-stream which no longer
uses the non-standard `__defineGetter__` syntax. By removing this, request can
now be loaded in IE10+ (which doesn't support `__defineGetter__`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "hawk": "~2.3.0",
     "aws-sign2": "~0.5.0",
     "stringstream": "~0.0.4",
-    "combined-stream": "~0.0.5",
+    "combined-stream": "~1.0.1",
     "isstream": "~0.1.1",
     "har-validator": "^1.6.1"
   },


simov:
Thanks :+1: what's the reason behind changing the version modifier? Can you change it back to `~`?

apechimp:
@simov just changed it back to `~`. I just did `npm install --save` :stuck_out_tongue_winking_eye:

dgsunesen:
I still have this problem? Updated request to latest version? Am I missing something?

dgsunesen:
It seems to be 'form-data' that has an older version of 'combined-stream' as a dependency.. version ~0.0.4 and that version of 'combined-stream' is dependant on an older version of 'delayed-stream'.. version 0.0.5 which have the non-standard ```__defineGetter__``` syntax

simov:
Thanks for the heads up, @dgsunesen, I'm planning on dropping the `form-data` dependency at some point in favor of our home made multipart module, as `form-data` doesn't seems to be actively maintained.

But in the meantime, in case you are sure that bumping the stream version there fixes your problem, you can open up a PR [there](https://github.com/felixge/node-form-data) and hope it gets published on NPM any time soon.

PR #1568, opened 2015-05-08
Fix stall when piping gzipped response

### LINKED ISSUE #1567: Weird issue when piping a chunked data from HTTP to a file in OS X
From [http://stackoverflow.com/questions/30085943/nodejs-unexpected-behaviour-when-piping-chunked-http-data-to-a-file-in-os-x?noredirect=1#comment48289848_30085943](http://stackoverflow.com/questions/30085943/nodejs-unexpected-behaviour-when-piping-chunked-http-data-to-a-file-in-os-x?noredirect=1#comment48289848_30085943).

I'm getting a weird issue when piping a chunked data from HTTP to a file in OS X.

When triggering the GET HTTP request that initiates the download, my server responds with the following headers:

    < content-disposition: attachment; filename="resin-Rpi2-0.1.0-0.0.12.img"
    < content-encoding: gzip
    < content-type: application/octet-stream
    < date: Wed, 06 May 2015 19:03:33 GMT
    < X-Frame-Options: DENY
    < x-powered-by: Express
    < transfer-encoding: chunked
    < Connection: keep-alive

My code goes like this:

    fs = require('fs')
    request = require('request')
    
    # If you want to try it by yourself, register to https://dashboard.resin.io
    # create a Raspberry Pi application (note the id on the url)
    # and get the token from the preferences page.
    APPID = '...'
    TOKEN = '...'
    DEST = 'imagetest.img'
    
    request
      url: "https://dashboard.resin.io/download?network=ethernet&wifiSsid=&wifiKey=&appId=#{APPID}"
      method: 'GET'
      gzip: true
      headers:
          Authorization: "Bearer #{TOKEN}"
    
    .on 'data', (data) ->
      console.log(data)
    
    .pipe(fs.createWriteStream(DEST))

If I run that script, I correctly get the initial chunks (corresponding to about 10MB of data) but the connection then suddenly ends, and no other event (close, end, etc) is triggered from any of the streams.

What's even more weird is that commenting out `gzip: true` leads to the desired behaviour (I of course get a *.img.gz file in that case), and it also works as expected if keeping the `gzip: true` and piping to `process.stdout` instead of to a file.

Also notice the issue is not reproducible on Ubuntu. It only seems to happen in OS X and Windows.

`cURL` also works as expected:

    $ curl -H "Authorization: Bearer [token]" "https://dashboard.resin.io/download?network=ethernet&wifiSsid=&wifiKey=&appId=[appid]" -o image.img

I've proceeded to analyse the packet communication ([you can get the pcap here][1]):

I can see my IP is suddenly sending a ACK/FIN while the server is still sending me chunks:

    Id = 2984
    Source = 192.168.0.16
    Destination = 52.4.102.97
    Captured Length = 66
    Packet Length = 66
    Protocol = TCP
    Date Received = 2015-05-06 18:11:36 +0000
    Time Delta = 10.27246999740601
    Information = 52264 -> 443 ([ACK, FIN], Seq=2137874087, Ack=3255859962, Win=10407)

Followed by a long loop of RST and ACK:

    ...
    192.168.0.16 -> 52.4.102.97, 52264 -> 443 RST
    52.4.102.97 -> 192.168.0.16, 443 -> 52264 ACK
    ...

And finally an agreed ACK, FIN from both sides.

After a lot of experimentation, the issue is gone if I write the chunks to the destination stream manually instead of piping:

    destination = fs.createWriteStream(DEST)

    request
      url: "https://dashboard.resin.io/download?network=ethernet&wifiSsid=&wifiKey=&appId=#{APPID}"
      method: 'GET'
      gzip: true
      headers:
          Authorization: "Bearer #{TOKEN}"
    
    .on 'data', (data) ->
      destination.write(data)
  
  [1]: http://filebin.net/5h4k576f6t/ResinImageDownloadIssue.pcap

kevinoid:
When the response content is piped through additional streams for decoding (e.g. for gzip decompression), pause and resume calls should be propagated to the last stream in the pipeline so that back pressure propagates correctly.

This avoids an issue where simultaneous back pressure from the content decoding stream and from a stream to which Request is piped could cause the response stream to get stuck waiting for a drain event on the content decoding stream which never occurs.  See #1567 for an example.

This commit also renames `dataStream` to `responseContent` to remedy my previous poor choice of name, since the name will be exposed on the Request instance it should be clearer and closer to the name used to refer to this data in the relevant RFCs.

Fixes #1567

### DIFF
--- a/request.js
+++ b/request.js
@@ -1120,38 +1120,40 @@ Request.prototype.onRequestResponse = function (response) {
       self._ended = true
     })
 
-    var dataStream
+    var responseContent
     if (self.gzip) {
       var contentEncoding = response.headers['content-encoding'] || 'identity'
       contentEncoding = contentEncoding.trim().toLowerCase()
 
       if (contentEncoding === 'gzip') {
-        dataStream = zlib.createGunzip()
-        response.pipe(dataStream)
+        responseContent = zlib.createGunzip()
+        response.pipe(responseContent)
       } else {
         // Since previous versions didn't check for Content-Encoding header,
         // ignore any invalid values to preserve backwards-compatibility
         if (contentEncoding !== 'identity') {
           debug('ignoring unrecognized Content-Encoding ' + contentEncoding)
         }
-        dataStream = response
+        responseContent = response
       }
     } else {
-      dataStream = response
+      responseContent = response
     }
 
     if (self.encoding) {
       if (self.dests.length !== 0) {
         console.error('Ignoring encoding parameter as this stream is being piped to another stream which makes the encoding option invalid.')
-      } else if (dataStream.setEncoding) {
-        dataStream.setEncoding(self.encoding)
+      } else if (responseContent.setEncoding) {
+        responseContent.setEncoding(self.encoding)
       } else {
         // Should only occur on node pre-v0.9.4 (joyent/node@9b5abe5) with
         // zlib streams.
         // If/When support for 0.9.4 is dropped, this should be unnecessary.
-        dataStream = dataStream.pipe(stringstream(self.encoding))
+        responseContent = responseContent.pipe(stringstream(self.encoding))
       }
     }
+
+    self.responseContent = responseContent
 
     self.emit('response', response)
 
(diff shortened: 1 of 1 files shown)


simov:
@kevinoid :+1:

kevinoid:
Thanks @simov!

PR #1570, opened 2015-05-09
Discovered tests that weren't properly running

seanstrom:
I came across some tests in the test-body.js that didn't seem to be actually asserting anything.
It seems that assertions are only being done if there is an expectBody property and if the value of the expectBody property is a Buffer, otherwise it won't run an assertion and just pass. To make matters even more complicated some of the tests don't give an expectBody property, so I'm not completely sure how to fix those tests. Just wanted to bring this to our attention.

### DIFF
--- a/tests/server.js
+++ b/tests/server.js
@@ -94,7 +94,7 @@ exports.createPostValidator = function (text, reqContentType) {
         assert.ok(~req.headers['content-type'].indexOf(reqContentType))
       }
       resp.writeHead(200, {'content-type':'text/plain'})
-      resp.write('OK')
+      resp.write(r)
       resp.end()
     })
   }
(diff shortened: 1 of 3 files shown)


seanstrom:
@simov @nylen @FredKSchott

simov:
The last few tests without `expectBody` are using `server.createPostValidator` which doesn't return anything useful, and instead makes the assertion on the server. So if you change the `resp.write('OK')` to `resp.write(r)` you'll receive the data back in the request's callback, like in the `createPostJSONValidator` below.

Changing It actually fails in a very few places, those that check for the OK response. It's better to fix it that way IMO.

seanstrom:
Yeah that makes sense.
Also, is there any behavior we want to cover when we do send a PUT or POST request?
It doesn't seem like there is but I was thinking we may want to test the behavior that will always happen when we submit a PUT or POST.

simov:
Part of the assertions are already made in `createPostValidator`, my point was only about returning back the data from the server, just in case anyone want's to use it in their tests, because most of the methods in `server` are working that way. Now for these specific tests in test-body, you'll probably have to check if it makes sense to define `expectBody` in some or any of the remaining tests.

seanstrom:
I changed the test helpers and tests to reflect what you mentioned.

simov:
:+1:

PR #1574, opened 2015-05-13
Refresh the oauth_nonce on redirect (#1573)

### LINKED ISSUE #1573: OAuth parameters are not refreshed following a 302 redirect
After following a 302 Redirect, it appears the subsequent request is signed using the same OAuth parameters as the original request.

simov:
- Cache the initial oauth options passed to request in _oauth.params
- On subsequent calls to init() use the cached _oauth.params
  to invoke the oauth params generation logic again

Fixes #1573

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -9,6 +9,7 @@ var qs = require('qs')
 
 function OAuth (request) {
   this.request = request
+  this.params = null
 }
 
 OAuth.prototype.buildParams = function (_oauth, uri, method, query, form, qsLib) {
@@ -90,6 +91,7 @@ OAuth.prototype.concatParams = function (oa, sep, wrap) {
 
 OAuth.prototype.onRequest = function (_oauth) {
   var self = this
+  self.params = _oauth
 
   var uri = self.request.uri || {}
     , method = self.request.method || ''
--- a/request.js
+++ b/request.js
@@ -628,6 +628,8 @@ Request.prototype.init = function (options) {
 
   if (options.oauth) {
     self.oauth(options.oauth)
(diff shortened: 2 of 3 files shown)


marshall007:
@simov I just noticed that since the subsequent calls to init() occur after the logic in [`lib/redirect.js#L135-L139`](https://github.com/request/request/blob/12ff8334d50df2b0e83d9e8342148a36cc922e06/lib/redirect.js#L135-L139) if you are redirected to an external host your OAuth credentials will be forwarded.

Here's a failing test case where we start at `http://localhost:6767` and get redirected to `http://127.0.0.1:6768/`, but the credentials are still set:

```js
tape('no credentials on external redirect', function(t) {
  var s1 = http.createServer(function (req, res) {
    res.writeHead(302, {location:'http://127.0.0.1:6768/'})
    res.end()
  })
  var s2 = http.createServer(function (req, res) {
    res.writeHead(200, {'content-type':'text/plain'})
    res.end()
  })
  s1.listen(6767, function () {
    s2.listen(6768, function () {
      request.get(
        { url: 'http://localhost:6767/'
        , oauth:
          { consumer_key: 'consumer_key'
          , consumer_secret: 'consumer_secret'
          , token: 'token'
          , token_secret: 'token_secret'
          }
          t.equal(err, null)
          t.equal(res.request.headers.Authorization, undefined)
          s1.close(function () {
            s2.close(function () {
              t.end()
            })
          })
        })
    })
  })
})
```

simov:
Good catch @marshall007 and thanks for the test :+1: 

Pushed your test and a fix for it.

Issue #29, opened 2011-05-05
Redirects w/ POST on HTTP 302/303/308: Found


EvanCarroll:
# BACKGROUND
## [rfc1945 HTTP 1.0](http://tools.ietf.org/html/rfc1945)

```
   302 Moved Temporarily

   The requested resource resides temporarily under a different URL.
   Since the redirection may be altered on occasion, the client should
   continue to use the Request-URI for future requests.

   The URL must be given by the Location field in the response. Unless
   it was a HEAD request, the Entity-Body of the response should
   contain a short note with a hyperlink to the new URI(s).

   If the 302 status code is received in response to a request using
   the POST method, the user agent must not automatically redirect the
   request unless it can be confirmed by the user, since this might
   change the conditions under which the request was issued.

       Note: When automatically redirecting a POST request after
       receiving a 302 status code, some existing user agents will
       erroneously change it into a GET request.
```
## [rfc2616 HTTP 1.1](http://tools.ietf.org/html/rfc2616)

```
10.3.3 302 Found

   The requested resource resides temporarily under a different URI.
   Since the redirection might be altered on occasion, the client SHOULD
   continue to use the Request-URI for future requests.  This response
   is only cacheable if indicated by a Cache-Control or Expires header
   field.

   The temporary URI SHOULD be given by the Location field in the
   response. Unless the request method was HEAD, the entity of the
   response SHOULD contain a short hypertext note with a hyperlink to
   the new URI(s).


   If the 302 status code is received in response to a request other
   than GET or HEAD, the user agent MUST NOT automatically redirect the
   request unless it can be confirmed by the user, since this might
   change the conditions under which the request was issued.

      Note: RFC 1945 and RFC 2068 specify that the client is not allowed
      to change the method on the redirected request.  However, most
      existing user agent implementations treat 302 as if it were a 303
      response, performing a GET on the Location field-value regardless
      of the original request method. The status codes 303 and 307 have
      been added for servers that wish to make unambiguously clear which
      kind of reaction is expected of the client.
```

So what happened is after the release of HTTP 1.0, prior to the advent of HTTP 1.1 which brought codes `303`, and `307` the spec read that the client MUST NOT redirect `POST` verbs temporarily. Unfortunately, there was no mechanism in the spec to redirect `POST` verbs -- slight oversight -- and the browsers were redirecting solely on the basis of `Location` header. It is not common place for `HTTP 302` to be legacy for the more explicit `HTTP 303`. This functionality should be permitted in Request too.

[...]

PR #1575, opened 2015-05-13
Test for forever-agent PR #29 - Enable socket reuse

mgorczyca:
Added as requested by @simov

### DIFF
--- a/tests/test-agent.js
+++ b/tests/test-agent.js
@@ -36,8 +36,8 @@ tape('should have one persistent socket for two POST requests', function(t) {
       })
     }
     else {
-        response.socket.destroy()
-        t.end()
+      response.socket.destroy()
+      t.end()
     }
   })
 })
--- a/tests/test-agent.js
+++ b/tests/test-agent.js
@@ -15,6 +15,35 @@ tape('setup', function(t) {
   })
 })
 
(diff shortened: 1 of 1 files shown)


mgorczyca:
Looks like the error is that the test failed. The forever-agent PR #29 is needed for the test to succeed.

simov:
@mgorczyca thanks for your input. However your test is failing only on 0.10, and it is passing with 0.12 and io.js without applying your fix from the forever-agent repo. The error you are seeing in TravisCI is that the next test is timing out, which is completely different issue. Just skip the next test `tape.skip` and run `tape tests/test-agent.js` with both 0.12 and iojs - you'll see that your test is passing, so I'm not sure what exactly it tests.

mgorczyca:
Thanks for your feedback @simov. I thought I had run my test in a pure 0.12 environment, and discovered one override to 0.10. I originally discovered the problem in 0.10 where the test still (properly) fails. Once I corrected my 0.12 environment, the test passes there for me, just as it does for you. We are planning to use 0.12 at this point, so we can close the issue and the PR if you agree that is best.

simov:
Closing in favor of https://github.com/request/request/pull/1671 - request now defaults to `http(s).Agent` in node 0.12+ when the `forever` option is used.

Issue #1583, opened 2015-05-17
Error: custom CA behind corporate proxy


mainakae:
Hello all and congratulations for this awesome library!.

I've been experiencing some issues passing custom CA certificates to request. I followed the documentation, setting `agentOptions.ca` successfully in an non-proxy environment, but then I moved the code to a proxied one (proxy configured via `$http_proxy`), and found that it didn't work any more. Te error that returned the previously working code was that it couldn't verify the CA. After some debugging and code inspection I've summed-up the following: 

1. using `options.agentOptions.ca`, which is the way stated in the documentation, will work in a non-proxied env, but not in a proxied one.
2. using `https.globalAgent.options.ca`, which is not documented, but is stated in many places (stackoverflow, forums...) as a feasible way of setting it, will work on non-proxied but won't in proxied again.
3. using `options.ca`, which is not documented (I had to find it inspecting code and later in some issue comments in github) WORKS IN BOTH, non-proxied and proxied environments.

The problem lies in this piece of code (or so I think)

```javascript
function constructTunnelOptions(request) {
  var proxy = request.proxy

  var tunnelOptions = {
    proxy : {
      host      : proxy.hostname,
      port      : +proxy.port,
      proxyAuth : proxy.auth,
      headers   : request.proxyHeaders
    },
    headers            : request.headers,
    ca                 : request.ca,
    cert               : request.cert,
    key                : request.key,
    passphrase         : request.passphrase,
    pfx                : request.pfx,
    ciphers            : request.ciphers,
    rejectUnauthorized : request.rejectUnauthorized,
    secureOptions      : request.secureOptions,
    secureProtocol     : request.secureProtocol
  }

  return tunnelOptions
}
```

In order to account for other possibles sources of CA information, this code should become something like:

```javascript
var tunnelOptions = {
//...
    ca                 : request.ca ||  agentOptions.ca || https.globalAgent.options.ca
//...
}
```

This should take the right order of priority for CAs (request > agentOptions > globalAgent). There's a problem, though: `agentOptions` is not accessible from `constructTunnelOptions`. Other alternative could be to use `request.agent` (`agentOptions` are used to construct that `.agent`) but then again, `request.agent` is constructed way after `constructTunnelOptions` is called, thus no luck there neither.

[...]

Issue #1490, opened 2015-03-17
Not obeying customAgent CA 


klinquist:
Hi,
It seems that although I have  "ca" file (buffer obj)  in a https customAgent, request is not obeying it and still giving me a "Self signed cert in certificate chain..." error.  I've tried passing it in the globalAgent object as well as to the pool...

My environment is an AWS private subnet with a Squid caching proxy performing transparent SSL Bumps.  I have generated a self-signed CA which is working perfectly with modules that use node's core https library (such as aws-sdk).

When I run:
If I run
`openssl s_client -showcerts -connect www.googleapis.com:443`
I get the error I expect:
`Verify return code: 19 (self signed certificate in certificate chain)`

However, as long as I specify my CA, 
`openssl s_client -showcerts -CAfile /etc/ssl/certs/myCA.pem -connect www.googleapis.com:443`
Open SSL shows
`Verify return code: 0 (ok)`

I'm passing this customAgent on to request:

```js
    var customAgent = new https.Agent({
        ca : require('fs').readFileSync(process.env.SSL_CA_CERT),
    });
```

and still get the self-signed cert error. `NODE_DEBUG=request` seems to show the ca buffer as the request is made, but I still get the cert error...

Thanks!

froatsnook:
I don't have any experience with playing with request and CAs, but how are you passing `customAgent` to `request`?  Have you tried using `agentOptions` as is done in this example from the README?

```
request.get({
    url: 'https://api.some-server.com/',
    agentOptions: {
        ca: fs.readFileSync('ca.cert.pem')
    }
});
```

klinquist:
Nevermind.  When declaring the customAgent, I was using  https.globalAgent.ca  instead of  https.globalAgent.options.ca.

mainakae:
Using agentOptions wouldn't solve the problem for Klinquist, I'm affraid. There's more than one way to set the custom CA: via options.agentOptions.ca or via options.ca. In case you are using an organization proxy, the agentOptions.ca will not work! (but options.ca will). So... there's one of two possibilities here: to fix documentation and remove all the agentOptions.ca stuff, or fix the code so that ca can be set while tunneling from agentOptions aswell.

mainakae:
Update: it's being discussed and ¿solved? in  #1289

Issue #1290, opened 2014-11-26
client side certificate doc mistake


OferE:
There is a mistake in the documentation on how to use client side certificate.
The cert, key and passphrase should be given to the "options" object and not to the agentOptions.

I also created a pull request for this:
https://github.com/request/request/pull/1289

I didn't touch the SSLV3 paramters inside the agentOptions, so please also verify if this is the correct place for it.

I hope this helps.

nylen:
Fixed by #1585.

PR #1289, opened 2014-11-26
fix the readme to correctly describe how to use client side certificate

OferE:
There was an error in the documentation. client side certificate options are not part of the agentOptions object.

PR #1585, opened 2015-05-17
Fixing documentation regarding TLS options (#1583)

mainakae:
Fixing the documentation to recommend using `options.ca`, `options.cer`, `options.key` and `options.passphrase` instead of `options.agentOptions.` homologues, to avoid trouble in proxied environments, as commented in #1583

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,7 +584,30 @@ Note: The `SOCKET` path is assumed to be absolute to the root of the host file s
 ## TLS/SSL Protocol
 
 TLS/SSL Protocol options, such as `cert`, `key` and `passphrase`, can be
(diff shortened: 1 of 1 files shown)


nylen:
This looks correct to me, and I think it will help address a number of other issues, as you said, #1289, #1490, #1290, #1229, #1236.  However, I have not been following `request` closely for some months, @simov what do you think?

simov:
I'm for merging it, as it adds additional information instead of replacing parts of the documentation like it was in the previous PR.

lalitkapoor:
:+1:

nylen:
Thanks @mainakae !!

mainakae:
Great!, Thank you!

PR #1603, opened 2015-05-26
Migrate to codecov

### LINKED ISSUE #1577: Cannot view source on Coveralls
When you try and view the source of a file on Coveralls you see the following error instead:

```
The owner of this repo needs to re-authorize with Github; their OAuth credentials are no
longer valid so the file cannot be pulled from the Github API.
```

https://coveralls.io/builds/2561958/source?filename=request.js

simov:
Fixes #1577 (hopefully)

 From what I'm reading coveralls no longer works with the public token for some reason. Codeclimate, codeship and now coveralls requires full read/write access to all of your public/private repos + the org ones = which is insane IMO.

Codecov on the other hand requires minimal set of permissions for public projects + it seems that they support reporting via the GitHub status API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ node_js:
   - "0.10"
 sudo: false
 
-after_script: "npm run test-cov && cat ./coverage/lcov.info | ./node_modules/codecov.io/bin/codecov.io.js"
+after_script: "npm run test-cov && cat ./coverage/lcov.info | codecov && cat ./coverage/lcov.info | coveralls"
 
 webhooks:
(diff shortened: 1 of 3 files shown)


simov:
This is how the coverage looks like https://codecov.io/github/request/request

simov:
Codecov turned out to be under heavy development, so I'm postponing this PR until this https://github.com/codecov/support/issues/12 issue get resolved.

stevepeak:
Thanks for the PR. I'll be working on codecov/support#12 today. I look forward to this merge!

simov:
Thanks, @stevepeak the GitHub's status API seems to be working :+1:

PR #1615, opened 2015-05-31
Replace '.client' with '.socket' as the former was deprecated in 2.2.0.

ChALkeR:
Actually, it's not only deprecated, checking it with `.hasOwnProperty('client')` is broken in 2.2.0, but that's going to be fixed in 2.2.1.

See https://github.com/nodejs/io.js/pull/1572 https://github.com/nodejs/io.js/issues/1850 [nodejs/io.js#1852/files](https://github.com/nodejs/io.js/pull/1852/files)

### DIFF
--- a/request.js
+++ b/request.js
@@ -1040,10 +1040,10 @@ Request.prototype.onRequestResponse = function (response) {
 
   // XXX This is different on 0.10, because SSL is strict by default
   if (self.httpModule === https &&
-      self.strictSSL && (!response.hasOwnProperty('client') ||
-      !response.client.authorized)) {
+      self.strictSSL && (!response.hasOwnProperty('socket') ||
+      !response.socket.authorized)) {
     debug('strict ssl error', self.uri.href)
-    var sslErr = response.hasOwnProperty('client') ? response.client.authorizationError : self.uri.href + ' does not support SSL'
+    var sslErr = response.hasOwnProperty('socket') ? response.socket.authorizationError : self.uri.href + ' does not support SSL'
     self.emit('error', new Error('SSL Error: ' + sslErr))
     return
   }
--- a/tests/ssl/ca/localhost.js
+++ b/tests/ssl/ca/localhost.js
@@ -21,9 +21,9 @@ https.request({ host: 'localhost'
               , agent: agent
               , ca: [ ca ]
               , path: '/' }, function (res) {
-  if (res.client.authorized) {
+  if (res.socket.authorized) {
     console.log('node test: OK')
   } else {
-    throw new Error(res.client.authorizationError)
+    throw new Error(res.socket.authorizationError)
   }
 }).end()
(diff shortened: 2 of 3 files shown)


ChALkeR:
Travis errors exactly because it has version 2.2.0 that runs `npm install` that in turn uses `request` that is atm not compatible with 2.2.0.

othiym23:
> that in turn uses request that is atm not compatible with 2.2.0

This is backwards – it's io.js that has introduced a breaking change in a minor version and is now incompatible with `request`, `npm`, and any other package that uses request's `strictSSL` checking.

This patch shouldn't be accepted until and unless it adds support for `.client` as well as `.socket`, because it's going to go from a situation where request (and npm) are broken for io.js 2.2.0 to a situation where request (and npm) are broken for io.js and Node.js _everywhere else_. While it's at it, somebody should figure out if using `.hasOwnProperty()` is _absolutely necessary_ for this to work and replace it with a simple property presence test otherwise – `.hasOwnProperty()` seems unnecessarily stringent and is part of what made this bug so severe and unpredictable in the first place.

thefourtheye:
> if using `.hasOwnProperty()` is absolutely necessary for this to work

Exactly what I had in my mind too.

simov:
Yep both properties should be supported, and that should be implemented by simply converting the property to the one or the other before it's been used for the first time, in any of the contexts where it's used.

ChALkeR:
@othiym23 `.socket` is there from nodejs 0.1.90.

ChALkeR:
> This is backwards – it's io.js that has introduced a breaking change in a minor version and is now incompatible with request, npm, and any other package that uses request's strictSSL checking.

That is indeed a bug and will be fixed in 2.2.1, as I already mentioned. But it would stil emit a deprecation warning.

simov:
@ChALkeR 

```js
res.client = res.client || res.socket
```

ChALkeR:
@simov Why? Just why? `.client` and `.socket` are there from nodejs version 0.1.90 and they were always the same object.

othiym23:
To summarize my discussion with @ChALkeR on IRC: my initial comment about just swapping the property names being insufficient was in error: since `client` is an alias for `connection`, the suggested change is fine and will fix `request` with `io.js@2.2.0`.

It shouldn't be necessary, though, because there's no compelling reason to have removed this property from the object in the first place (and any further discussion along those lines should probably be had over in nodejs/io.js anyway, as this point isn't really related to `request`). As such I'm +0 on this PR.

simov:
Thanks for pointing that out @othiym23 I was wondering if I'm missing out something, while I was catching up with the issue.

ChALkeR:
The deprecation was completely reverted in 2.2.1. When I was filing this PR, I assumed that the deprecation would be fixed instead (as https://github.com/nodejs/io.js/pull/1852 was doing at that moment).

With this, there is actual reason to merge this, except for using a documented property instead of a non-documented one. If you want to merge this, I could change the commit message. If not, just close it.

simov:
doh I just saw that request is a bundled dependency in npm .. anyway 2.57 is out

ChALkeR:
Thanks.

PR #1624, opened 2015-06-02
Fix readme Table of Contents links

RebootJeff:


simov:
Thanks for the input, @RebootJeff, unfortunately your fix breaks the header links when rendered in GitHub. I filed an issue in the NPM's web site repository https://github.com/npm/newww/issues/1021

RebootJeff:
Oh bummer. It didn't occur to me that GitHub and NPM would render the readme differently. Thanks for the feedback, @simov.

PR #1638, opened 2015-06-11
Use the `extend` module to deep extend when using the defaults method

simov:
@mikeal @nylen I'm introducing the [extend](https://www.npmjs.com/package/extend) module into core. This fixes the flat extend behavior when using `util._extend` in `defaults` for any other object key that the `headers`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@
 
 'use strict'
 
-var extend                = require('util')._extend
+var extend                = require('extend')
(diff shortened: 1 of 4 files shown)


Rykka:
when using npm install request (2.6.4), it throws cannot find module extend,after install it manually (3.0) , it works ok. don't know this is a error caused by npm or something

PR #1646, opened 2015-06-17
Clarify the nature of setting `ca` in `agentOptions`

jeffcharles:
The `ca` option in https://nodejs.org/api/tls.html#tls_tls_connect_options_callback indicates that the certificate presented must be signed by one of the certificates provided. The existing wording in request led me to believe that setting it augmented the list of acceptable certificates rather than replacing it.

This caused a problem when we switched from a testing environment which used a self-signed certificate to a staging environment which used a third-party signed certificate.


simov:
Thanks :+1:

PR #1652, opened 2015-06-22
Update README.md

daniel347x:
The documentation for the 'encoding : null' argument does not emphasize that this is necessary for binary response data, something that is not obvious - see http://stackoverflow.com/a/14855016/368896. I've added a parenthetical comment to emphasize this in the documentation. This would have saved me (and hopefully will save others).

### DIFF
--- a/README.md
+++ b/README.md
@@ -766,7 +766,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 
 ---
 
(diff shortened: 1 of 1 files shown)


simov:
Yep, I agree that `encoding:null` is a too clever option. However I would like the wording to be a bit more declarative, and especially I don't like the _Warning_ bits. How about something a bit shorter and less noisy?

> **Note:** if you expect binary data, you should set `encoding: null`

Binary data can be anything so mentioning PDF and images is useless. In fact this option is related to the next one as well (`gzip`), in case you may want to receive the gzipped content and handle it by yourself.

daniel347x:
That's good!  I've updated it to match.  I did leave it in parentheses, to give the sense that this is a 'heads up'.  The note would have saved me!

simov:
Thanks :+1:

Issue #1669, opened 2015-07-07
Documentation missing for `agent` option


triccardi-systran:
`request` accepts a http/https Agent as `agent` option, but it's not documented.
What is documented is how to pass a pool of agents as `pool` option, but the code first checks if `self.agent` exists before using `self.pool` to get the http agent:

https://github.com/request/request/blob/2ca06330c9e8d438b07c9efd9a077303d2e59ec8/request.js#L486-L490

This should be documented & tested, or it should be removed... Otherwise we cannot rely on it.

simov:
I agree there are other options missing from the docs as well, like `agentClass` and `agentOptions` mentioned here #1637

simov:
@triccardi-systran see https://github.com/request/request/issues/1671

tvdstaaij:
It sure seems to work, I just passed in a `https.Agent` object with the `keepAlive` option set to `true` (Node 0.12) and verified using an external TCP analysis tool that the connections are kept alive and reused, while this is not the case when the custom agent is not passed in.

This is a very useful feature but it took me over an hour of googling and analyzing source code to find it, so seconding the documentation request. I also suggest updating the documentation for `pool` so that the differences and use cases for the two are made clear.

Issue #1637, opened 2015-06-11
`forever: true` causes memory to leak like the deepwater horizon


devTristan:
```javascript
var request = require('request');

var concurrency = 128;
var inProgress = 0;
var startedCount = 0;
var finishedCount = 0;

function fillQueue() {
  while (inProgress < concurrency) {
    makeRequest();
  }
}

function makeRequest() {
  inProgress++;
  startedCount++;

  request.post({
    uri: 'http://localhost/',
    forever: true
  }, function(){
    inProgress--;
    finishedCount++;
    fillQueue();
  });
}

fillQueue();

setInterval(function(){
  console.log(inProgress+' requests in progress, '+startedCount+' started, '+finishedCount+' finished');
}, 1000);
```

When executed, this script will leak memory until you run out of it. This effect has been observed on ubuntu 14.04 using node 0.11.16 and 0.12.4. **There is no memory leak in 0.10.25**.

Don't run this script if you have a server running locally on port 80, this might DOS it.

Commenting out `forever: true` will cause this script to run as normal, but without the memory leak, and presumably without keepalive support.

seamusabshere:
@devTristan do you know what is going on?

devTristan:
My first guess would be something in [here](https://github.com/request/forever-agent/blob/master/index.js) preventing request data from being garbage-collected. Very possible that I'm wrong, though.

devTristan:
@simov gentle ping

mikeal:
0.12+ makes this feature totally unnecessary. We should check if the node version is greater than 0.10.x and print a warning when using this feature and silently turn it off.

simov:
:+1: I was just trying to figure out what's the difference between forever-agent and the current Agent implementation in core.

simov:
@devTristan, theoretically this should give you the same results, without the memory leak. Let me know what do you think.

```js
var request = require('request');
var http = require('http');

var concurrency = 128;
var inProgress = 0;
var startedCount = 0;
var finishedCount = 0;

function fillQueue() {
  while (inProgress < concurrency) {
    makeRequest();
  }
}

function makeRequest() {
  inProgress++;
  startedCount++;

  request.post({
    uri: 'http://localhost/',
    agentOptions: {keepAlive:true, maxSockets:1},
    agentClass: http.Agent
    // forever: true
  }, function(){
    inProgress--;
    finishedCount++;
    fillQueue();
  });
}

fillQueue();

setInterval(function(){
[...]

PR #1671, opened 2015-07-10
Add tests and docs for using the agent, agentClass and agentOptions. Forever option defaults to using http(s).Agent in node 0.12+

simov:
Fixes https://github.com/request/request/issues/1637
Fixes https://github.com/request/request/issues/1669

Feedback: @mikeal @devTristan @triccardi-systran and anyone else interested in

- Take a look at the agent tests [here](https://github.com/simov/request/blob/agent-docs/tests/test-agent.js). Just a bunch of simple use cases, also note that the `forever` option defaults to `http(s).Agent` in node 0.12+
- Take a look at the [docs](https://github.com/simov/request/tree/agent-docs#requestoptions-callback) as well. A little bit down below there is a section for `agent` related options. It would be great if you read through all of them carefully and let me know if I'm not getting something right.

### DIFF
--- a/README.md
+++ b/README.md
@@ -772,7 +772,11 @@ The first argument can be either a `url` or an `options` object. The only requir
 
 ---
 
-- `pool` - An object describing which agents to use for the request. If this option is omitted the request will use the global agent (as long as [your options allow for it](request.js#L747)). Otherwise, request will search the pool for your custom agent. If no custom agent is found, a new agent will be created and added to the pool.
+- `agent` - `http(s).Agent` instance to use
+- `agentClass` - alternatively specify your agent's class name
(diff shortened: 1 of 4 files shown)


simov:
@mikeal @devTristan @triccardi-systran (+ anyone else) any feedback on this one?

devTristan:
I don't love `forever` being 0.10 only, as the docs say, but the code clearly supports the option in more recent versions. Why not get rid of the `console.warn` and keep `forever` as an option available for all versions?

simov:
Probably a more appropriate way to document the `forever` option would be to say that it defaults to `http.Agent` in node 0.12+ which was suggested by @mikeal, also that's the way to get rid of potential bugs such as the memory leak one in your issue here https://github.com/request/request/issues/1637. The `console.warn` can be removed in that case I agree.

simov:
@devTristan updated the [docs](https://github.com/simov/request/tree/agent-docs#requestoptions-callback) and removed the warning.

PR #1686, opened 2015-07-20
Fix #1685: setHeader() with undefined value

### LINKED ISSUE #1685: self.setHeader(self._form.getHeaders()) has no effect due to 'undefined' value
The ```setHeader``` on [request.js, L576](https://github.com/request/request/blob/master/request.js#L576) has no effect due to its ```undefined``` value (due to [caseless, index.js, L52](https://github.com/request/caseless/blob/master/index.js#L52)).

We can keep the ```undefined``` value condition, but we must set a value in order to set the headers.

[PR #1686](https://github.com/request/request/pull/1686)

garymathews:
- Allow ```setHeader``` to work by not using an ```undefined``` value

[Issue #1685](https://github.com/request/request/issues/1685)

### DIFF
--- a/request.js
+++ b/request.js
@@ -573,7 +573,7 @@ Request.prototype.init = function (options) {
 
     if (self._form && !self.hasHeader('content-length')) {
       // Before ending the request, we had to compute the length of the whole form, asyncly
-      self.setHeader(self._form.getHeaders())
+      self.setHeader(self._form.getHeaders(), true)
       self._form.getLength(function (err, length) {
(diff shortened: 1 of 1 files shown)


simov:
Thanks for finding the right place for the fix, @garymathews, your commit is pushed here https://github.com/request/request/pull/1687. I'll publish it shortly.

simov:
[2.60](https://github.com/request/request/issues/1688) is published.

PR #1700, opened 2015-07-27
debug() when JSON.parse() on a response body fails

phillipj:
Previously JSON.parse() failures was silently ignored.

Although `response.body` would still contain the invalid JSON response body as a string, instead of a JS object, it's valueable to know the JSON was invalid while debugging.

I was tempted to `.emit('error', new Error('Invalid JSON received'))` but that might be a little harse.

Any thoughts?


simov:
Yep, seems good. Emitting error there won't be good as there are plenty of consumers that handle this case by themselves already.

Issue #1660, opened 2015-06-26
Ability to distinguish connect timeouts and read timeouts


kevinburkeshyp:
This request times out because the `connect` socket call fails:

```javascript
var request = require('request');
request.get('http://google.com:81', {timeout: 3000}, function(err, response, body) {
    console.log(err.code === 'ETIMEDOUT');
    process.exit(0);
});
```

This request times out because the `read` socket call doesn't see any response in 3 seconds:

```javascript
var request = require('request');
request.get('http://localhost:5501', {timeout: 3000}, function(err, response, body) {
    console.log(err.code === 'ETIMEDOUT');
    process.exit(1);
})
```

(I'm using [hamms](https://github.com/kevinburke/hamms) in the second test)

It would be nice to be able to distinguish these. Particularly, requests to a third party provider are failing and it would be good to distinguish whether the request actually made it to their server (their problem) or not (our problem).

It would also be neat because you could figure out whether or not the request made it to the server - a connection timeout is always retry-able, a read timeout on a POST is not.

matteocontrini:
Up

simov:
Fixed in https://github.com/request/request/issues/1711

PR #1711, opened 2015-08-03
Add ability to detect connect timeouts

kevinburke:
It's valuable to be able to detect whether a timeout occurred on the `connect`
syscall or on the `write`/`read` syscall of an HTTP request - the former
implies that your client can't reach the server and the latter implies the
server might not have responded.

For backwards compatibility, and standardization with Unix errno codes, the
`code` property of connect timeout errors must remain as 'ETIMEDOUT'. Instead
of changing this value, add a new `connect` property on timeout errors, which
is set to true if the error occurred on the `connect` attempt. Clients have
to opt in to checking it - in the common read timeout case, it will be set to
`false`.

Updates the documentation around each timeout to clarify what is going on, and
what situations can trigger each timeout.

Updates the README with more accurate documentation about timeout times
in the wild. I tried `time curl http://10.255.255.1` and `time curl
http://google.com:81` on four different Linux machines (Heroku, boot2docker,
Centos on DigitalOcean, a VPS) and the timeout times ranged from 60 seconds to
120 seconds, much longer than the 20 seconds previously suggested.

Updates the README with information on how to detect connection timeout
failures.

### DIFF
--- a/request.js
+++ b/request.js
@@ -804,21 +804,33 @@ Request.prototype.start = function () {
 
   if (self.timeout && !self.timeoutTimer) {
     var timeout = self.timeout < 0 ? 0 : self.timeout
+    // Set a timeout in memory - this block will throw if the server takes more
+    // than `timeout` to write the HTTP status and headers (corresponding to
+    // the on('response') event on the client). NB: this measures wall-clock
+    // time, not the time between bytes sent by the server.
     self.timeoutTimer = setTimeout(function () {
+      var connectTimeout = self.req.socket && self.req.socket.readable === false
       self.abort()
       var e = new Error('ETIMEDOUT')
       e.code = 'ETIMEDOUT'
+      e.connect = connectTimeout
       self.emit('error', e)
     }, timeout)
 
-    // Set additional timeout on socket - in case if remote
-    // server freeze after sending headers
     if (self.req.setTimeout) { // only works on node 0.6+
+      // Set an additional timeout on the socket, via the `setsockopt` syscall.
+      // This timeout sets the amount of time to wait *between* bytes sent
+      // from the server, and may or may not correspond to the wall-clock time
+      // elapsed from the start of the request.
+      //
+      // In particular, it's useful for erroring if the server fails to send
+      // data halfway through streaming a response.
       self.req.setTimeout(timeout, function () {
         if (self.req) {
           self.req.abort()
           var e = new Error('ESOCKETTIMEDOUT')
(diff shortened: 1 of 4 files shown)


kevinburke:
Fixes https://github.com/request/request/issues/1660.

request.js
```diff
@@ -804,21 +804,32 @@ Request.prototype.start = function () {
 
   if (self.timeout && !self.timeoutTimer) {
     var timeout = self.timeout < 0 ? 0 : self.timeout
+    // Set a timeout in memory - this block will throw if the server takes more
+    // than `timeout` to write the HTTP status and headers (corresponding to
+    // the on('response') event on the client). NB: this measures wall-clock
+    // time, not the time between bytes sent by the server.
     self.timeoutTimer = setTimeout(function () {
+      var connectTimeout = self.req.socket && self.req.socket.readable === false;
```
kevinburke: Need to read this value before aborting, since the socket gets destroyed. Could pull the error creation up here too, not sure which way is preferable.

tests/test-timeout.js
```diff
@@ -43,6 +43,19 @@ if (process.env.TRAVIS === 'true') {
     })
   })
 
+  tape('should set connect to false', function(t) {
```
kevinburke: Note Travis skips the tests in this file. I verified the tests passed on my local machine (a Mac). Would appreciate testing in other environments, especially Windows - I'm not sure if Node has the same socket API on Windows machines.

kevinburke:
Whoops - made a breaking change when writing a comment and didn't see the Travis results. Looks like there are some failures related to expired certificates. Not sure if they are related or not; I can investigate further.

simov:
For the certificates just run `tests/ssl/ca/gen-client.sh` and `tests/ssl/ca/gen-localhost.sh`

kevinburke:
> Also add npm-debug.log to .gitignore and remove it from your commit.

Whoops - not sure how that snuck in there. Sorry for wasting your time. Should be fixed now.

kevinburke:
Cert failures should hopefully go away once https://github.com/request/request/pull/1712 is merged. I will rebase once that's in.

kevinburke:
Tests are back to green. The code coverage on this is 0, because the timeout tests don't run on Travis. I believe that they should run okay, but let me know what you think, or if there's a workaround I can use for code coverage here.

simov:
Looks good to me, thanks for working on and documenting this one, @kevinburke! I'll leave this PR open for a while in case anyone else have something to add.

PR #1712, opened 2015-08-03
Set certificate expiration to August 2, 2018

kevinburke:
Updates the gen-* scripts to take a `-days` argument, and set that `days`
argument to 3 years from now. I believe without this argument the certificates
default to a 30 day expiry. Code lifted from
https://github.com/shazow/urllib3/blob/master/dummyserver/certs/README.rst.

Generates new SSL certs with a long expiry.


kevinburke:
whoops... need to check if this passes tests. 1 minute.

simov:
Something isn't right with our tests lately, it's on my TODO list to investigate why they timeout so badly.

kevinburke:
When I opened the PR the branches hung on `npm install` (not sure why). Triggered a rebuild and it looks like the tests are all passing.

simov:
Thanks, now you can pull master :+1:

Issue #1709, opened 2015-07-31
"forever" option is useless in node 0.12.*


calibr:
because it's creates new agents every request, for example:
```js
var request = Request.forever({
  maxSockets: 2
});

var r1 = request("http://google.com", function() {});
var r2 = request("http://google.com", function() {});

console.log(r1.agent === r2.agent); //prints false, but should be true
```
Problem is here https://github.com/request/request/blob/master/request.js#L486 .

This can be fixed by replacing this line by something like this:
```js
self.agentClass = self.httpModule.Agent;
self.agentOptions.keepAlive = true;
```

Such behaviour makes sockets not reusable, because creates new pool(agent) every http request.

simov:
Generally speaking the `forever` should not be used in anything higher than node 0.10 (as stated in the docs), see https://github.com/request/request/pull/1671 for details. Your fix proposal looks good :+1:

PR #1715, opened 2015-08-05
Fix forever option in node > 0.10 #1709

calibr:
Fixes https://github.com/request/request/issues/1709

Makes `forever` option work fine in node > 0.10, so now it doesn't create new agent every request, and reuse old agents. Issue https://github.com/request/request/issues/1709.


calibr:
hm test fails on fresh iojs v3, but it also fails on master branch with this io version on my local computer...

simov:
Looking good to me :+1:

PR #1716, opened 2015-08-05
Do not create Buffer from Object in setContentLength(iojs v3.0 issue)

calibr:
Due to changes in io.js v3.0 creating Buffer from Object throws an Error. This PR just prevents this


simov:
:+1:

PR #1721, opened 2015-08-09
update README.md

arbaaz:
uri is not defined.


simov:
Hi, @arbaaz. I honestly don't see what is this PR fixing. Feel free to elaborate.

arbaaz:
When I used the same snippet in my code exception occurred('uri is not defined'). When I changed to `url` it started working.

simov:
Now I see it:

```js
var url = 'http://www.google.com';
```

is defined in the code block from above. Is that it?

arbaaz:
Yes. that is it.

PR #1736, opened 2015-08-18
Reuse request agent on redirect

### LINKED ISSUE #1304: POST 301 redirects don't respect "agent" property

joshux:
Fixes #1304
Closes https://github.com/request/request/pull/1451

The problems I found before the fix: 
1. _updateProtocol was incorrect and is redundant with init(). 
2. The code was mixing configuration (options.agentClass) with state (self.agentClass). In my opinion, it would be better if the configuration is immutable which shouldn't be deleted on redirection.

### DIFF
--- a/lib/redirect.js
+++ b/lib/redirect.js
@@ -120,7 +120,6 @@ Redirect.prototype.onResponse = function (response) {
   // request.method = 'GET' // Force all redirects to use GET || commented out fixes #215
   delete request.src
   delete request.req
-  delete request.agent
   delete request._started
   if (response.statusCode !== 401 && response.statusCode !== 307) {
     // Remove parameters from the previous response, unless this is the second request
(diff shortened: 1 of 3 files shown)


Freyert:
I like the cut of your jib @joshux, that's brilliant to delineate configuration and state like that.

simov:
Thanks for solving this up @joshux, I really like the clean up part :+1: 

I just removed the options thing here https://github.com/request/request/commit/84a5ea37cfda1f2a11c12c73b4ed3c06679f4cc1

PR #1743, opened 2015-08-20
Fix options not being initialized in defaults method

simov:
Fixes https://github.com/request/request/commit/ae403dcf92e196991dc1f90a275b794fbf44d260#commitcomment-12808144


calibr:
:+1:

PR #1747, opened 2015-08-26
Add more descriptive error message on ETIMEDOUT

ruimarinho:
This makes the `ETIMEDOUT` error more useful and inline with what io.js throws.

### DIFF
--- a/request.js
+++ b/request.js
@@ -815,7 +815,7 @@ Request.prototype.start = function () {
     self.timeoutTimer = setTimeout(function () {
       var connectTimeout = self.req.socket && self.req.socket.readable === false
       self.abort()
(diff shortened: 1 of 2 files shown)


FredKSchott:
I agree that this would be better with a helpful error message, but it will definitely break anyone checking that the message equals `ETIMEDOUT` or `ESOCKETTIMEDOUT` instead of checking the code. We would need to message this as a breaking change.

A good first step would be to add the requested url as a property of the error object. That way we could at least support new users who need this info, without breaking old users. @ruimarinho could you create a new PR (based off latest master) to add this?

ruimarinho:
Done. https://github.com/request/request/pull/2551.

PR #1750, opened 2015-08-29
Revert doc about installation of tough-cookie added in #884

LoicMahieu:
I merged PR #884 but it was a mistake. `tough-cookie` is builtin and required by `request` for a while. There is no need for user to install it manually. So I revert the PR. Sorry for confusion.


18nleung:
Thanks for clarifying!

PR #1767, opened 2015-09-10
Query strings now cooperate with unix sockets

### LINKED ISSUE #1766: Using options.qs with a unix socket fails
When the following request is done:
```js
request({
    uri: 'http://unix:/var/run/docker.sock:/containers/json'
})
```
[Some juggling is done](https://github.com/request/request/blob/master/request.js#L258) on `request.uri` to make the unix socket work. `request.uri.host` becomes `/var/run/docker.sock`.

However, when a query string is provided by the user, for example: 
```js
request({
    uri: 'http://unix:/var/run/docker.sock:/containers/json',
    qs: { all: 'true' }
})
```
Parsing the query string [clobbers the unix socket](https://github.com/request/request/blob/master/request.js#L1162). `request.uri.host` reverts to `unix`.

JoshWillik:
Fixes #1766

### DIFF
--- a/request.js
+++ b/request.js
@@ -256,17 +256,7 @@ Request.prototype.init = function (options) {
 
   // Support Unix Sockets
   if (self.uri.host === 'unix') {
-    // Get the socket & request paths from the URL
-    var unixParts = self.uri.path.split(':')
-      , host = unixParts[0]
-      , path = unixParts[1]
-    // Apply unix properties to request
-    self.socketPath = host
-    self.uri.pathname = path
-    self.uri.path = path
-    self.uri.host = host
-    self.uri.hostname = host
-    self.uri.isUnix = true
+    self.enableUnixSocket()
   }
 
   if (self.strictSSL === false) {
@@ -1163,6 +1153,10 @@ Request.prototype.qs = function (q, clobber) {
   self.url = self.uri
   self.path = self.uri.path
 
+  if ( self.uri.host === 'unix' ) {
+    self.enableUnixSocket()
+  }
+
   return self
 }
 Request.prototype.form = function (form) {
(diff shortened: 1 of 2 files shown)


tests/test-unix.js
```diff
@@ -27,7 +31,12 @@ tape('setup', function(t) {
 })
 
 tape('unix socket connection', function(t) {
```
simov: I would like to have two separate tests, the original one without querystring and a new one with querystring. That way if someone removes the unix bits from the init function the tests will throw an error.

request.js
```diff
@@ -1411,6 +1405,19 @@ Request.prototype.destroy = function () {
     self.response.destroy()
   }
 }
+Request.prototype.enableUnixSocket = function () {
```
simov: Not the best place to put this function. Notice the comment above - *Streams API*

request.js
```diff
@@ -1163,6 +1153,10 @@ Request.prototype.qs = function (q, clobber) {
   self.url = self.uri
   self.path = self.uri.path
 
+  if ( self.uri.host === 'unix' ) {
```
simov: Minor style issue - remove spacing.

simov:
Hey, @JoshWillik nice work on this one. I added a few comments.

request.js
simov: Also not sure if that needs to be defined on the prototype as a public method, probably a function that accepts an `uri` object will be just fine.

JoshWillik: @simov Line 1249 sets `this.socketPath`, which requires access to `this`, no?

JoshWillik: I've fixed all the things you highlighted, but I can make it a seperate by doing `fn (uri, context){}` or something if you prefer.

simov:
:+1:

JoshWillik:
:D Thanks @simov

JoshWillik:
@simov How often does request update the NPM package? I'd like to be able to stop doing my own query string assembly in my code :stuck_out_tongue:

simov:
Usually when someone asks about it, or when there is a bugfix that affects a lot of people. Strictly non semver :) but that'll change at some point. I'll let you know when the new release is published on NPM.

simov:
@JoshWillik 2.62 is published.

JoshWillik:
Hooray! Thank you :D

PR #1772, opened 2015-09-15
Set default application/json content-type when using json option

### LINKED ISSUE #1759: `json: true` doesn't set Content-Type
Documentation for the `json` options says:
> json - sets body but to JSON representation of value and adds Content-type: application/json header. Additionally, parses the response body as JSON.

This doesn't work for `json: true`:
```js
var x = request({
  uri: 'http://whatever',
  json: true
})
// FIXME this should include `content-type: application/json`
console.log(x.headers)
// > { host: 'whatever', accept: 'application/json' }
```
The actual implementation also checks for `self.body`:
```js
if (typeof val === 'boolean') {
  if (self.body !== undefined && self.getHeader('content-type') !== 'application/x-www-form-urlencoded') {
    self.body = safeStringify(self.body)
    if (!self.hasHeader('content-type')) {
      self.setHeader('content-type', 'application/json')
    }
  }
}
```
I'm dealing [with an API](https://mesosphere.github.io/marathon/docs/rest-api.html#delete-v2-groups-groupid) that expects a `Content-Type: application/json` even for a DELETE request without a body, so the existing implementation doesn't work here. I'm working around that by explicitly setting the header on every `json: true` request, of which there are many littered through this application.

I suggest changing the implementation to make the `!self.hasHeader('content-type')` check independent of the `self.body` check, so if `val` is a boolean, try to set the header, ignoring the missing body.

If that's acceptable, I can look into creating a patch, at least the implementation change should be straight forward.

jzaefferer:
Will set the header when using `json: true` without a body.

Fixes #1759

### DIFF
--- a/request.js
+++ b/request.js
@@ -1198,6 +1198,9 @@ Request.prototype.json = function (val) {
   }
 
   self._json = true
+  if (!self.hasHeader('content-type')) {
+    self.setHeader('content-type', 'application/json')
+  }
   if (typeof val === 'boolean') {
     if (self.body !== undefined) {
       if (!/^application\/x-www-form-urlencoded\b/.test(self.getHeader('content-type'))) {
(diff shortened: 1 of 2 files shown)


simov:
:+1:

jzaefferer:
Could you also tag and publish this? Thanks.

simov:
Done.

PR #1811, opened 2015-10-03
Enable loose cookie parsing in tough-cookie

Sebmaster:
tough-cookie 2.1.0 got a loose cookie mode which is more how browsers actually behave. It implies an empty key for value-only cookies.

Not sure how request handles strict spec-compliance vs. practicality, so I thought I'd just get this started with a PR.

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -13,13 +13,13 @@ exports.parse = function(str) {
   if (typeof str !== 'string') {
     throw new Error('The cookie function only accepts STRING as param')
   }
-  return Cookie.parse(str)
+  return Cookie.parse(str, {loose: true})
 }
 
 // Adapt the sometimes-Async api of tough.CookieJar to our requirements
 function RequestJar(store) {
   var self = this
-  self._jar = new CookieJar(store)
+  self._jar = new CookieJar(store, {looseMode: true})
 }
 RequestJar.prototype.setCookie = function(cookieOrStr, uri, options) {
   var self = this
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "node-uuid": "1.4.3",
     "qs": "5.1.0",
     "tunnel-agent": "0.4.1",
-    "tough-cookie": "2.1.0",
+    "tough-cookie": "2.2.0",
     "http-signature": "0.11.0",
     "oauth-sign": "0.8.0",
     "hawk": "3.1.0",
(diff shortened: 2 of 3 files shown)


simov:
Thanks, @Sebmaster, I just subscribed to the tough-cookie's PR thread.

@lalitkapoor want to chime in?

stash-sfdc:
tough-cookie@2.2.0 now has the two PRs by @Sebmaster, specifically the changes to `Cookie.parse` and `new CookieJar`.

Tangental, but FYI: tough-cookie@2.1.0 updated the embedded public suffix list. Even if you don't want the loose parsing mode, `request` should probably update to depend on `tough-cookie >= 2.1.0`

lalitkapoor:
thanks @stash-sfdc I'll review these shortly and create another pr for updating tough-cookie.

lalitkapoor:
https://github.com/request/request/pull/1824

lalitkapoor:
This makes sense to me. We want to be reasonably flexible, and since most browsers are lenient in this manner. By default, request should be as well.

In the future it would be nice to eventually be able to specify these sort of things in a request option somewhere (in case someone does want to be strict).

:+1: Thanks for the pr @Sebmaster :)

Sebmaster:
After thinking this through a bit more with @domenic, it seems like providing `{loose: true}` explicitly to the `setCookie` method is a bad idea since it'll cause inconsistencies across the eco-system.

Instead, I'd suggest only enabling `looseMode` for wrapper cookie jars and if cookie jars are passed in from another source, just go with their specified behaviour. That means that users will have to specify `looseMode: true` in most cases themselves, until tough-cookie bumps the major and (hopefully) enables that behaviour by default. However, this solution will at least not produce any inconsistencies.

lalitkapoor:
@Sebmaster Just for reference I'm leaving a note to also look at your comments here - https://github.com/tmpvar/jsdom/pull/1255

Sebmaster:
Updated the PR to not force stuff on the cookie jar.

lalitkapoor:
@Sebmaster is this reviewable again? And what did you change from before? Removed the loose option in setCookie? Thanks.

simov:
@lalitkapoor that was my thinking, @Sebmaster probably it would be better to keep all commits in PRs at least until prior merging, otherwise the conversation seems broken, and you can't diff between the changes.

Sebmaster:
Sorry about that. Only removed the `loose: true` on setCookie and instead put it on the cookie jar.

lalitkapoor:
:+1:

Issue #1817, opened 2015-10-05
Proposal - `Request#readResponseBody`


pvoisin:
There's a cool piece of code [here](https://github.com/request/request/blob/master/request.js#L1016:L1065) that one could benefit from if it were extracted and exposed as a function of `Request#prototype`... :)

If I find some time to do it I'll definitely enjoy contributing. In the meantime I'm just dropping those lines for any feedback.

simov:
I agree :+1:

pvoisin:
What about [that](https://github.com/request/request/pull/1818)?

simov:
Closed via #1818

PR #1818, opened 2015-10-05
Extract `readResponseBody` method out of `onRequestResponse`

pvoisin:
As mentionned in issue #1817 I've extracted the portion of `Request#onRequestResponse` that reads the response's body into `Request#readResponseBody`.


simov:
Thanks :+1: 
Closes #1817

PR #1819, opened 2015-10-06
Run stringify once

mgenereu:
Code seems to be running _qs.stringify twice for no reason.


simov:
Good catch @mgenereu :+1:

PR #1820, opened 2015-10-06
Set href as request.js uses it

mgenereu:
If href was never defined in an original, dynamically generated request options object the rest of the code referencing href fails.


simov:
That's great, but can you add a simple test for it? Probably [test-isUrl](https://github.com/request/request/blob/master/tests/test-isUrl.js) is a good place for that.

mgenereu:
I'm on it.

mgenereu:
Test provides the qs and no href.  It breaks before fix was applied.

simov:
:+1:

PR #1821, opened 2015-10-06
Implement support for RFC 2617 MD5-sess algorithm.

BigDSK:
Current implementation does not handle the RFC 2617 MD5-sess algorithm. I've added that ability along with a simple test. I've also confirmed this is working against an IIS server ( the only real-world test I have  at my disposal ).

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -50,8 +50,6 @@ Auth.prototype.bearer = function (bearer, sendImmediately) {
 
 Auth.prototype.digest = function (method, path, authHeader) {
   // TODO: More complete implementation of RFC 2617.
-  //   - check challenge.algorithm
-  //   - support algorithm="MD5-sess"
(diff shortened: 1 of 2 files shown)


simov:
Looking good, thanks for the contribution @BigDSK :+1:

simov:
@BigDSK do you think it's a good idea to update the TODO list [here](https://github.com/BigDSK/request/blob/53caeb21eaf018dab643216f56811cdf9d2248ba/lib/auth.js#L52-L62)?

BigDSK:
@simov Good idea. I had missed the TODO list. 

Overall my understanding of RFC 2617 is that we now cover both options. If your understanding is that algorithm may be something other than "MD5" or "MD5-sess", then we should incorporate some type of error handling. Silently falling back to a default/incorrect algorithm was confusing in my case.

I believe we're good to go though.

PR #1835, opened 2015-10-12
req.body undefined or null same logic branch

stonephp:
I found the problem when use frisby.

### DIFF
--- a/lib/querystring.js
+++ b/lib/querystring.js
@@ -41,6 +41,9 @@ Querystring.prototype.parse = function (str) {
 }
 
 Querystring.prototype.rfc3986 = function (str) {
+  if (typeof str !== 'string') {
+    return str;
+  }
   return str.replace(/[!'()*]/g, function (c) {
     return '%' + c.charCodeAt(0).toString(16).toUpperCase()
   })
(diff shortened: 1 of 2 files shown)


simov:
This breaks our tests because `null` is allowed value for JSON encoded request body. Can you elaborate on what exactly your use case is?

stonephp:
I trace the code, found ./request.js 1212 line, self._qs.rfc3986(self.body).
```js
Querystring.prototype.rfc3986 = function (str) {
   return str.replace(/[!'()*]/g, function (c) {
```

Here, str.replace throw error, when str is null.

simov:
Yep, I know that it throws there. Can you provide a short code example of your options setup that you are passing to request?

stonephp:
```js
request = require('request');

request({
  json: true,
  uri: 'http://www.google.com/',
  body: null,
  method: 'GET',
  headers: {
    'content-type': 'application/x-www-form-urlencoded'
  }
});
```

simov:
Oh, the `body:null` thing .. got it! You have to remove the semicolons `;` in your last commit - TravisCI is failing again.

fritx:
image

Same here..  request@2.79.0

PR #1861, opened 2015-10-24
Remove redundant call to Stream constructor

watson:
I'm not sure if this is some sort of witchcraft that I'm not familiar with, but I would normally not call the parent constructor twice as it have been done now for about a year since ef4016f.

If this is indeed on purpose I think it's one of those rare occasions that deserves a code-comment explaining it :wink:


simov:
:+1:

Issue #1872, opened 2015-10-28
Har support - mimeType and charset


sod:
Chrome 46, Debugger tools Har export.

I tried to repeat a request via har data.

My har file has: `req.postData.mimeType === 'application/x-www-form-urlencoded; charset=UTF-8'` but the switch `switch (data.postData.mimeType) {` in https://github.com/request/request/blob/master/lib/har.js#L63 does not expect to have the `; charset=UTF-8` in here.

The request itself works. But if I modify the postData, it is not recognised as no switch/case matches.

simov:
@sod check out this PR https://github.com/request/request/pull/1876

PR #1876, opened 2015-10-29
Implement loose matching for har mime types

simov:
Fixes https://github.com/request/request/issues/1872

### DIFF
--- a/lib/har.js
+++ b/lib/har.js
@@ -60,43 +60,53 @@ Har.prototype.prep = function (data) {
   }
 
   // prep body
-  switch (data.postData.mimeType) {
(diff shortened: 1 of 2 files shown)


sod:
Thank you for taking over the issue.

There is another `switch/case` in: https://github.com/request/request/pull/1876/files?w=1#diff-13c8aeb1b94f67f8c33c1cab8c4b1021R165

And maybe add a `; charset=UTF-8` to one of the mime types in https://github.com/request/request/blob/master/tests/fixtures/har.json and see if the tests pass :)

simov:
Yep, these are good suggestions. Pushed a fix.

PR #1880, opened 2015-11-01
Error on initialization should prevent actual HTTP request from being sent

arcana261:
Hello

Working on a propitiatory piece of app the other day, I ran into a strange exception happening on nextTick. This exception would complain of **can not call method request of undefined**. When I inspected the problem deeply I understood the problem was that the URI which I passed to **request** did not have any protocol defined, like **/bad-uri** but also the problem seemed to be triggered because a stream was piped to **request** to upload some files to a WebDAV server. I also confirmed that **request** did in fact call my callback with provided exception error but was still trying to make the actual HTTP request regardless. The exception happened because initialization was failed thus leaving **undefined** variables behind but this case of error was never checked in **Request.write** method which was called in nextTick following stream of data from my file. So I tried to fix this problem by:

1. Creating a test case (which needed [tape-catch](https://www.npmjs.com/package/tape-catch) to fail tests creating exceptions on process level)
2. Setting **self._aborted** to **true** on every error inside **request**
3. Checking **self._aborted** in methods **Request.write** and **Request.end**
4. Fixing a test case in **test-bearer-auth** that falsely relied on **request** making HTTP requests regardless of errors in it's initialization

The following is an image showing the stacktrace of exception which I'm talking about:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "server-destroy": "^1.0.1",
     "tape": "^4.2.0",
     "taper": "^0.4.0",
-    "bluebird": "^3.0.2"
+    "bluebird": "^3.0.2",
+    "tape-catch": "^1.0.4"
   }
 }
--- a/request.js
+++ b/request.js
@@ -196,7 +196,10 @@ Request.prototype.init = function (options) {
       self._callbackCalled = true
       self._callback.apply(self, arguments)
     }
-    self.on('error', self.callback.bind())
+    self.on('error', function () {
+      self._aborted = true
+      return self.callback.bind().apply(this, arguments)
+    })
     self.on('complete', self.callback.bind(self, null))
   }
 
@@ -1333,20 +1336,24 @@ Request.prototype.pipe = function (dest, opts) {
 }
 Request.prototype.write = function () {
   var self = this
-  if (!self._started) {
-    self.start()
+  if (!self._aborted) {
+    if (!self._started) {
+      self.start()
+    }
+    return self.req.write.apply(self.req, arguments)
   }
-  return self.req.write.apply(self.req, arguments)
 }
 Request.prototype.end = function (chunk) {
   var self = this
-  if (chunk) {
-    self.write(chunk)
+  if (!self._aborted) {
+    if (chunk) {
+      self.write(chunk)
+    }
+    if (!self._started) {
+      self.start()
+    }
+    self.req.end()
   }
-  if (!self._started) {
-    self.start()
-  }
-  self.req.end()
 }
 Request.prototype.pause = function () {
   var self = this
--- a/tests/test-bearer-auth.js
+++ b/tests/test-bearer-auth.js
@@ -169,7 +169,7 @@ tape('null bearer', function(t) {
     }
   }, function(error, res, body) {
     t.equal(res.statusCode, 401)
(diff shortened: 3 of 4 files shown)


simov:
Hey, @arcana261, very good work on this PR :+1: 

However I'm proposing a slightly simpler solution here: https://github.com/simov/request/commit/6931cea23ca5cfe3817dc8bb809ecb249921fa3c

Basically you need only the `self.abort()` call on line 284, then I just changed the check in `write` and `end` to be slightly more readable `if (self._aborted) {return}`.

Unfortunately even this doesn't cover all possible cases, and our error handling is kind of awkward due to the way the module is being initialized. You can read about it here: https://github.com/request/request/pull/1558

Still I think that aboring on line 284 is slightly better for this particular case, because it says `fatal error` anyway. Your solution is not bad either, but it's going to work only if you have a callback, which unfortunately is not always the case.

arcana261:
Dear good Sir @simov. Sorry for replying so late... (holidays in my country!)

I think it is smart... I'm just sorry for missing the ```abort``` method :smile: Regarding the callback, I would suggest we take a default callback if none was provided that throws an exception or just simply prints out in the console. Still I have two questions regarding your simov@6931cea23ca5cfe3817dc8bb809ecb249921fa3c 

1. Why would you replace-back ```tape``` instead of ```tape-catch```. Because the test won't fail using just ```tape```
2. ```t.equal(numBearerRequests, 13)``` fails in ```auth-bearer``` test because the HTTP request was not sent due to aborting it

I would definitely say this will not cover all our errors but I think we are making an improvement here :+1: Of course we are not expected to put a big ```try-catch``` on whole initialization anyway haha!  I think that taking a default callback sounds more fun to me! (I do this all the time!) and I just hope I will be of help in finding more bugs in the future :smile:

simov:
Providing a callback is optional, and having one means that the response will be buffered into memory - not something that you want to have by default.

Also I don't want to have the `_abort` flag set in the error event handler. Because that flag is used in slightly different context, that's why I prefer calling the `abort` method explicitly for that type of error. And having the error handler for the callback case won't work when you are streaming.

No idea why you used the `tape-catch` module in the first place. Search for `t.throws` in the test suite to see how you can test code that throws.

As for the `auth-bearer` test - like I said, our error handling is far from perfect at the moment. With my change the request is being aborted only when the URI is invalid, that's why this test passes.

Happy holidays! :tada:

arcana261:
Ahh I see... Thank you for your patient response. So we need to ```abort()``` the error in which that test generates maybe? and BTW I used the ```tape-catch``` because the error is actually generated in ```nextTick``` which I think ```throws``` would not be able to detect.

Thanks! :blush:

simov:
re: `nextTick` that makes sense.

> So we need to abort() the error in which that test generates maybe?

I don't understand your question.

What I'm suggesting is to merge my last commit on top of yours. Quoting the relevant bits from my previous comment:

> Also I don't want to have the _abort flag set in the error event handler. Because that flag is used in slightly different context, that's why I prefer calling the abort method explicitly for that type of error. And having the error handler for the callback case won't work when you are streaming.

arcana261:
Ok... Thanks! I appreciate it :smile:

simov:
:+1:

PR #1881, opened 2015-11-01
Fix tunneling after redirection from https

falms:
Current stable module cannot handle https-to-http redirection with proxy properly.

When redirected from https to **http** with `options.tunnel = undefined`,
the **http** request should be proxied without CONNECT tunneling.

This pull-request includes these changes:

1. use `options` again after redirection.
2. fix test case for https-to-http redirection.
3. fix description about `tunnel` option.

This pull-request solves [this problem](https://github.com/atom/apm/issues/322).

### DIFF
--- a/lib/redirect.js
+++ b/lib/redirect.js
@@ -135,6 +135,8 @@ Redirect.prototype.onResponse = function (response) {
         // changing ports or protocols).  This matches the behavior of curl:
         // https://github.com/bagder/curl/blob/6beb0eee/lib/http.c#L710
         request.removeHeader('authorization')
+        delete request.options.auth
+        delete request.options.oauth
       }
     }
   }
@@ -145,7 +147,7 @@ Redirect.prototype.onResponse = function (response) {
 
   request.emit('redirect')
 
-  request.init()
+  request.init(request.options)
 
   return true
 }
--- a/request.js
+++ b/request.js
@@ -137,6 +137,7 @@ function Request (options) {
   self._redirect = new Redirect(self)
   self._tunnel = new Tunnel(self)
   self.init(options)
+  self.options = options
 }
 
 util.inherits(Request, stream.Stream)
--- a/README.md
+++ b/README.md
@@ -799,8 +799,7 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
 - `tunnel` - controls the behavior of
   [HTTP `CONNECT` tunneling](https://en.wikipedia.org/wiki/HTTP_tunnel#HTTP_CONNECT_tunneling)
   as follows:
-   - `undefined` (default) - `true` if the destination is `https` or a previous
-     request in the redirect chain used a tunneling proxy, `false` otherwise
+   - `undefined` (default) - `true` if the destination is `https`, `false` otherwise
    - `true` - always tunnel to the destination by making a `CONNECT` request to
      the proxy
    - `false` - request the destination as a `GET` request.
--- a/lib/tunnel.js
+++ b/lib/tunnel.js
@@ -113,14 +113,7 @@ function Tunnel (request) {
 
 Tunnel.prototype.isEnabled = function (options) {
   var request = this.request
-  // Tunnel HTTPS by default, or if a previous request in the redirect chain
-  // was tunneled.  Allow the user to override this setting.
-
-  // If self.tunnel is already set (because this is a redirect), use the
-  // existing value.
-  if (typeof request.tunnel !== 'undefined') {
-    return request.tunnel
-  }
+  // Tunnel HTTPS by default. Allow the user to override this setting.
 
   // If options.tunnel is set (the user specified a value), use it.
   if (typeof options.tunnel !== 'undefined') {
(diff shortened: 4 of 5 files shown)


simov:
I don't feel comfortable altering that logic. Why do you think that the current implementation is wrong?

I would like @nylen to chime in here as well.

falms:
Thanks for your review!

There are some reasons:

1. [HTTP CONNECT tunneling](https://en.wikipedia.org/wiki/HTTP_tunnel#HTTP_CONNECT_tunneling) describes only about first https request. It is not about http request after redirection.
2. Major useragents (i.e. IE, Chrome, Firefox) just GET http url via proxy without tunneling even if it is after redirection from https url.
3. Current `tunnel = undefined` is not compatible with major/default squid configurations. Squid without `SSL_ports port 80` denies `CONNECT server:80`.

For these reasons, I changed tunneling logic.

nylen:
I spent a good 20 minutes looking through the history of this code.  As far as I can tell, in #1374 I was just trying to preserve the existing behavior of the library, which was [a bit convoluted](https://github.com/request/request/blob/3a2c38e38ce1c050b2229bbf6bd6f701f74daf9f/request.js#L296).

I don't have a problem with the change to the tunneling logic given that (1) existing browsers behave this new way and (2) this behavior is very well-tested.

Leaving some inline notes about changes that look unrelated to the tunneling behavior.

lib/redirect.js
```diff
@@ -135,6 +135,8 @@ Redirect.prototype.onResponse = function (response) {
         // changing ports or protocols).  This matches the behavior of curl:
         // https://github.com/bagder/curl/blob/6beb0eee/lib/http.c#L710
         request.removeHeader('authorization')
+        delete request.options.auth
+        delete request.options.oauth
```
nylen: This does not appear related to the tunneling changes, can you explain?

lib/redirect.js
```diff
@@ -145,7 +147,7 @@ Redirect.prototype.onResponse = function (response) {
 
   request.emit('redirect')
 
-  request.init()
+  request.init(request.options)
```
nylen: This makes me nervous.

request.js
```diff
@@ -137,6 +137,7 @@ function Request (options) {
   self._redirect = new Redirect(self)
   self._tunnel = new Tunnel(self)
   self.init(options)
+  self.options = options
```
nylen: This has a bit of a bad code smell because we already set lots of properties on `self` and want to be careful about adding more without good naming and clear reasons.

simov:
@nylen yep, those lines you commented are probably not going in, my question was about the tunneling logic only. I'll take a look at the issues you mentioned :+1:

lib/redirect.js
falms: I think that re-calling `init` with original `options` is better. Well, is this bad idea?

simov: @falms it might be better, but given the existing code base it's better to keep things consistent instead of adding even more noise.

lib/redirect.js
falms: You are right.. This is related to `init` changes in redirect.js, not tunneling.

falms:
Thanks for your reviews, @simov @nylen!

Here is another way. Please see this commit.
https://github.com/falms/request/commit/f4e9b0b95cd714ce564c5c61dfda8b266d5cd1bb

If this is better, I will commit this to fix-tunneling-after-redirection branch too.

simov:
@falms take a look at this commit https://github.com/simov/request/commit/7943dce24aaf0b5ecb85148ad8e8c045addc4ed4 My thinking is that the `tunnelOverride` flag initialization happens only once, and the right place for it is in the Tunnel's constructor.

Now, it's a bit weird that the `request` argument in the constructor holds the initial user options as well, but that's just how request is currently implemented.

falms:
@simov Holding and re-using initial user options is not so good as I thought a few days ago. I want to merge https://github.com/simov/request/commit/7943dce24aaf0b5ecb85148ad8e8c045addc4ed4 into `falms:fix-tunneling-after-redirection`.

Could you tell me the best way to merge, please?

simov:
Check out your branch `fix-tunneling-after-redirection-simple` containing your last proposal and pull mine:

```
git pull git@github.com:simov/request.git fix-tunneling-after-redirection-simple
```

Then you can push it into another PR - I think that's the easiest way.

nylen:
I prefer the first version of this change (https://github.com/falms/request/commit/d678bfc14e7170546abfdbf9b0e3967711e04839) over https://github.com/falms/request/commit/f4e9b0b95cd714ce564c5c61dfda8b266d5cd1bb - it's a simpler change that doesn't require adding an extra property.

simov:
@nylen the extra state property is needed, because you want the check in `isEnabled` to be executed each time. The old code was checking for the `tunnel` state first, which always is the last one used - that's the fix.

Take a look at the final version of the [function](https://github.com/simov/request/blob/7943dce24aaf0b5ecb85148ad8e8c045addc4ed4/lib/tunnel.js#L117-L134) and the [initialization](https://github.com/simov/request/blob/7943dce24aaf0b5ecb85148ad8e8c045addc4ed4/lib/tunnel.js#L112-L114) this is what we are talking about.

falms:
I created new PR #1894.

Well, both versions have extra property.
first version: [`request.options`](https://github.com/falms/request/blob/d678bfc14e7170546abfdbf9b0e3967711e04839/request.js#L140)
final version: [`tunnel.tunnelOverride`](https://github.com/simov/request/blob/7943dce24aaf0b5ecb85148ad8e8c045addc4ed4/lib/tunnel.js#L113)

In addition, changing `init` requires additional modifications unrelated to tunneling like `delete request.options.auth`.
Therefore, I pushed safer version.

PR #1889, opened 2015-11-05
Update qs dependency to 6.0.0

idolize:
Update to the latest `qs` version.

Fixes the following error when using the latest version of `qs` in addition to `request`:

```
npm WARN unmet dependency /node_modules/jsdom/node_modules/request requires qs@'~5.2.0' but will load
npm WARN unmet dependency /node_modules/qs,
npm WARN unmet dependency which is version 6.0.0
```


simov:
On hold until we drop support for everything below 4.0.0 https://twitter.com/eranhammer/status/662013345550065664

PR #1894, opened 2015-11-07
Fix tunneling after redirection from https (Original: #1881)

falms:
Current stable module cannot handle https-to-http redirection with proxy properly.

When redirected from https to **http** with `options.tunnel = undefined`,
the **http** request should be proxied without CONNECT tunneling.

This pull-request includes these changes:

1. use initial `options.tunnel` again after redirection.
2. fix test case for https-to-http redirection.
3. fix description about `tunnel` option.

Closes #1881


simov:
Thanks @falms :+1:

PR #1895, opened 2015-11-09
Fix missing curly bracket

harianus:
In the `README.md` file


simov:
No curly bracket is needed there. Take a look at the `url` property before the `form` - it's part of the options object.

harianus:
Oeps, sorry!

PR #1952, opened 2015-12-08
Adds example for Tor proxy

prometheansacrifice:


simov:
That's interesting. We have an [examples](https://github.com/request/request/tree/master/examples) folder so probably it's better to put that there.

prometheansacrifice:
@simov It IS from the example directory!

Are you suggesting I put it in a separate `.js` file?

simov:
Hmm right, I saw `README.md` and I though it's the main readme.

PR #1955, opened 2015-12-09
Check if the functions exist before executing

rik-smit:
Sometimes you get the error self.req.abort/response.resume is not a function. To prevent the error first check if the function exists.

### DIFF
--- a/request.js
+++ b/request.js
@@ -839,7 +839,7 @@ Request.prototype.onRequestResponse = function (response) {
   }
   if (self._aborted) {
     debug('aborted', self.uri.href)
-    response.resume()
+    if(typeof response.resume == 'function') response.resume();
     return
   }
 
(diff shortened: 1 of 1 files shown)


simov:
I want a bit more background on this, can you provide a short code snippet to reproduce both cases?

FredKSchott:
Related to https://github.com/request/request/issues/2315

The problem is that `response` and `self.req` are not valid request/response objects when they should be. Adding checks on the function call stops that error, but doesn't fix the overall problem that the function doesn't exist. 

Can you share easy steps to reproduce this? That could help us find out why `response` and `self.req` aren't what we expect.

FredKSchott:
Closing as abandoned (no activity in past year). 

Please reopen a new PR based off of master with passing tests if you are still interested in getting this merged and I'll be happy to take a look.

Issue #1949, opened 2015-12-08
helpers.isReadStream identifies only fs-created streams


jackwanders:
https://github.com/request/request/blob/master/lib/helpers.js#L41

`.isReadStream` checks to see if the provided object has three truthy properties: `.readable`, `.path`, and `.mode`. While these three properties all exist on streams created by [`fs.createReadStream`](https://nodejs.org/api/fs.html#fs_fs_createreadstream_path_options), this is not true for all readable streams that may contain file data.

In particular, I am working on a project using `sails@0.11.3` and `skipper@0.5.8`. The "upstream" paradigm that `skipper` uses creates a transform stream containing file data. This allows `sails` users to write file management middleware that sends file data to any desired storage location without first writing any data to disk. Because the file data is coming directly from the request and into the transform stream, no file system read is happening, and therefore no `.path` and `.mode` properties exist for these streams which contain file data.

I'm opening this issue to ask whether there is a more flexible way for `request` to identify readable streams. Node exposes four different stream classes that implement the readable interface, `Readable`, `Duplex`, `Transform`, and `PassThrough`. All of these stream types appear to share two properties:

```javascript
['Readable', 'Transform', 'Duplex', 'PassThrough'].forEach(function(type) {
  var stream = new require('stream')[type]();
  console.log(type, !!stream._readableState, !!stream.readable);
});

/* Output
Readable true true
Transform true true
Duplex true true
PassThrough true true
*/
```

One possible refactor that appears to allow `request` to accept non fs file streams while avoiding a regression is:

```javascript
function isReadStream (rs) {
  return rs.readable && ((rs.path && rs.mode) || rs._readableState);
}
```

Are there any foreseeable downsides to opening `request` up to accepting any readable stream, as opposed to only those created by `fs`? Is there a catastrophic failure point somewhere where `request` is assuming the stream to be an `fs` stream?

jackwanders:
Another option would be to check to see if the provided `rs` object is an instance of `stream.Stream`:

```javascript
var stream = require('stream');
['Readable', 'Duplex', 'Transform', 'PassThrough', 'Writable'].forEach(function(type) {
  var rs = stream[type]();
  console.log(type, rs.readable, rs instanceof stream.Stream);
});

/* Output
Readable true true
Duplex true true
Transform true true
PassThrough true true
Writable undefined true
*/
```

simov:
Hi @jackwanders thanks for the thorough bug report!

[...]

PR #1964, opened 2015-12-11
Use isstream for more robust stream.Readable detection

jackwanders:
See https://github.com/request/request/issues/1949 for the original issue description.

In short, `helpers.isReadStream` is overly restrictive in what it deems to be a readable stream. Specifically, only streams created via `fs` that stream file data from disk pass the test of this function. However, it is entirely valid to have a non `fs` stream that contains valid file data (e.g. streams generated from file data contained in `multipart/form-data` requests.

This pull request updates `helpers.isReadStream` to use the `.Readable` method contained within the existing dependency `isstream`. However, `isstream.isReadable` on its own is not restrictive enough to detect the kinds of readable streams that `request` needs to detect, as the `http.IncomingMessage` and its sub-classes are also readable streams, yet they need to be rejected by our test.

So, the combination of `isstream.isReadable` and `instanceof http.IncomingMessage` should be sufficient to successfully detect non http readable streams that may contain file data.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,7 +1,9 @@
 'use strict'
 
 var jsonSafeStringify = require('json-stringify-safe')
+  , isReadable = require('isstream').isReadable
   , crypto = require('crypto')
+  , http = require('http')
 
 function deferMethod() {
   if (typeof setImmediate === 'undefined') {
@@ -38,8 +40,8 @@ function md5 (str) {
   return crypto.createHash('md5').update(str).digest('hex')
 }
 
-function isReadStream (rs) {
-  return rs.readable && rs.path && rs.mode
+function isReadStream (obj) {
+  return isReadable(obj) && !(obj instanceof http.IncomingMessage)
 }
 
 function toBase64 (str) {


jackwanders:
Hmm, while the tests pass, this change appears to break with `FormData` streams when I test it in my app. 

I'll debug and come back when I know more.

jackwanders:
Well, upon further debugging, it appears that the original problem encountered in #1949 is not, in fact, a bug within `request`. It turns out to be a side effect of the `form-data` package and how it determines the length of streams when building `multipart/form-data` objects.

In short, the lack of a `.path` property on file streams built by `skipper` causes this return statement to fire without the `FormData.prototype._trackLength` function ever calling `next()`, which in turn caues the entire request to fail.

While I believe that there may be an improvement to `request` in refining the `isReadStream` helper, I no longer believe it's the bug that I originally suspected it to be.

jackwanders:
I'm closing this PR as it's no longer valuable in my mind. To wrap up my personal story here, it turns out that `request` and `form-data` already support my use-case; I had to specify the `formData.file.options.knownLength` property with the `byteCount` value of the stream created by `skipper`. Once I added this value, everything works as expected.

donpark:
I found this issue belatedly because `multipart/form-data` requests were failing with S3 stream although both `Content-Length` and `Content-Type` were exposed via CORS configuration.

To workaround `form-data`'s seemingly unnecessary check for `path`, I set a bogus `path` value like this:

    let stream; // stream from s3.getObject
    stream.path = 'bogus/path/for/form-data';

It's a nutty hack but works and *seemingly* has no side-effect. Hope `form-data` removes unnecessary path check soon.

PR #1966, opened 2015-12-14
Add support for SPNEGO/Kerberos authentication

rustyconover:
Add optional dependency on the kerberos module

Add examples and documentation to README.md

Signed-off-by: Rusty Conover rusty@twosigma.com


simov:
That's great, but we'll definitely need tests for it. Do you think it will be possible to create a build script for it on Linux and Mac systems? Also there have to be a way to deploy it on TravisCI as well.

rustyconover:
@simov do we have the ability to setup an entire Kerberos infrastructure (KDC included) and join the client to that realm to run the tests?  I'll check with some of my sources to see what it would entail, but I believe it is non trivial.

simov:
I personally don't have experience with Kerberos. I'm assuming you are on Windows, but I noticed that the author of the `kerberos` module mentions something about:

> Furthermore the kerberos module requires the MIT Kerberos package to correctly compile on UNIX operating systems. Consult your UNIX operation system package manager what libraries to install.

I'm willing to invest some time in this, I just wanted to ask you first for some guidance in case you have some experience with running Kerberos on Linux or something.

rustyconover:
Hi @simov,

I wouldn't say I have a lot of experience with Kerberos, but [Two Sigma](http://www.twosigma.com), where I work does.  We use it for almost everything, so everyone who works here does know a bit about it.  I have tested this code using Linux and the MIT Kerberos package, I haven't tested it on Windows.  Do you have a Windows Kerberos environment?

Rusty

simov:
Check out this test https://github.com/simov/request/commit/9249e048377dd6f567ee9bd4bfa8bd303d52b513

These are the modules that I installed on my Ubuntu box:

```bash
$ sudo apt-get install krb5-kdc krb5-admin-server libkrb5-dev
```

And then there is a small portion of configuration needed:

```bash
# add principal
$ sudo kadmin.local -q "addprinc -randkey HTTP/localhost"
# create keytab
$ sudo kadmin.local -q "ktadd -k /etc/krb5.keytab HTTP/localhost"
# in order to access the keytab store from non root user (like the node process)
$ sudo chmod 777 /etc/krb5.keytab
```

So these are the commands that will be executed in TravisCI as well, before running the test suite. The only problem is that the above packages have interactive prompt during install, so I'm not sure about that. Also the kerberos tests will be enabled only in TravisCI environment because currently we don't use docker to run our tests.

I'll probably add more stuff to my branch, and a few code comments here as well.

README.md
```diff
@@ -273,6 +273,14 @@ request.get('http://some.server.com/', {
     'bearer': 'bearerToken'
   }
 });
+// or with SPNEGO/Kerberos
+request.get('http://some.server.com/').auth(null, null, false, null, true);
```
simov: That's just insane. We should probably not bother documenting it at all.

README.md
```diff
@@ -273,6 +273,14 @@ request.get('http://some.server.com/', {
...
+// or
+request.get('http://some.server.com/', {
+  'auth': {
+    'negotiate': true
```
simov: As far as I understand there are other authenticating mechanisms such as [NTLM](https://en.wikipedia.org/wiki/NT_LAN_Manager) for example, so `negotiate: true` might not be the most approprite configuration to use. Will investigate further these days.

lib/auth.js
```diff
@@ -48,6 +55,66 @@ Auth.prototype.bearer = function (bearer, sendImmediately) {
   }
 }
 
+
+Auth.prototype.negotiate = function(request, authHeader, sendImmediately) {
+  var self = this
+
+  if (kerberos === undefined) {
```
simov: That's probably not needed, as this method looks public, but in fact it's used through internal `_auth` instance, and you are checking for undefined inside the `onRequest` method.

lib/auth.js
```diff
@@ -48,6 +55,66 @@ Auth.prototype.bearer = function (bearer, sendImmediately) {
...
+    // Nothing to do if we don't have Kerberos
+    // or a current credential cache
+    return
+  }
+
+  self.hasAuth = true
+
+  if (!sendImmediately) {
```
simov: This should be:

```js
if (sendImmediately === false) {return}
```

By default `sendImmediately` should be `true`, including for `undefined`, that's how it's documented. And btw the test that I pushed won't work with `if (!sendImmediately)`

lib/auth.js
```diff
@@ -48,6 +55,66 @@ Auth.prototype.bearer = function (bearer, sendImmediately) {
...
+
+        var content = context.response
+
+        k.authGSSClientClean(context, function(err) {
+          if (err) {
+            self.request.emit('error', new Error('auth() error with Kerberos: ' + err))
+            return
+          }
+
+          request.setHeader('Authorization', 'Negotiate ' + content)
```
simov: The `authorization` header should be set outside, like the rest of the auth schemes in this file. The error should be returned as well.

lib/auth.js
```diff
@@ -48,6 +55,66 @@ Auth.prototype.bearer = function (bearer, sendImmediately) {
...
+        })
+        return
+      })
+      return
+    })
+
+  // Return a fake authorization header that will
+  // be replaced by the real header later on when it
+  // is available
+  return 'pending'
```
simov: This doesn't do anything.

lib/auth.js
```diff
@@ -124,15 +191,16 @@ Auth.prototype.digest = function (method, path, authHeader) {
...
     , request = self.request
-
   var authHeader
-  if (bearer === undefined && user === undefined) {
+  if (bearer === undefined && user === undefined && negotiate === undefined) {
     self.request.emit('error', new Error('no auth mechanism defined'))
   } else if (bearer !== undefined) {
     authHeader = self.bearer(bearer, sendImmediately)
+  } else if (negotiate && kerberos !== undefined) {
+    self.negotiate(request, undefined, sendImmediately)
```
simov: How `authHeader` is supposed to be used inside the `kerberos` method if you are passing `undefined always?

package.json
```diff
@@ -43,6 +43,9 @@
     "is-typedarray": "~1.0.0",
     "har-validator": "~2.0.2"
   },
+  "optionalDependencies": {
```
simov: Can't do that as npm2 installs the optional dependencies by default. Specifying the dependency in the docs should be enough.

request.js
```diff
@@ -539,6 +543,10 @@ Request.prototype.init = function (options) {
     }
 
     var end = function () {
+      if (self._auth.pendingAuth) {
+        self.once('authReady', end)
```
simov: :+1:

README.md
rustyconover: I agree, the current auth() function signature isn't the most extensible especially if we want to add additional methods.

lib/auth.js
rustyconover: I don't really like the idea of `sendImmediately` being set to true. 

If a server's hostname isn't in a KDC's registered list you're going to get a GSSAPI error saying that the server is unknown.  Ideally we'd get the 401 response the first time and present the challenge data to generate the header, if we don't get a 401 we don't need to provide the authentication.  Not all servers use Kerberos, but we want the ability to authenticate to the ones that do.

By keeping the default to false, we can work with what each server requires.

lib/auth.js
rustyconover: The Kerberos auth information is returned asynchronously, it is impossible to set it in the same way as we're not returning a value back.  But I'm happy to make this match your style.

lib/auth.js
rustyconover: That's right, will remove.

lib/auth.js
rustyconover: This function will never have the SPNEGO challenge data (its returned by the 401 response), hence we always initialize the SPNEGO data with `undefined` here.  This will only be used if `sendImmediately` is set to true.

lib/auth.js
simov: That might be true, I just wanted to point it out, it's not the most urgent thing.

lib/auth.js
simov: TBH I'm not sure how the challenge works yet. Can you pull the test that I provided and probably try to add another one using a challenge? Also what configuration should I use for Kerberos on my workstation. I'll probably look at it tomorrow.

lib/auth.js
simov: More tests are needed :)

simov: That probably makes sense. I guess `sendImmediately` is set to true by default for basic auth because you are supplying your `user` name and `pass` so it kind of makes sense to send them by default.

> Not all servers use Kerberos, but we want the ability to authenticate to the ones that do.
> By keeping the default to false, we can work with what each server requires.

I'm still not aware what can go so horribly wrong by having `sendImmediately` set to true by default, but anyway, I just added a test case for that https://github.com/simov/request/blob/feature-kerberos-auth/tests/test-kerberos.js#L55-L62

lib/auth.js
simov: Yep, totally forgot about the `onResponse` call.

lib/auth.js
rustyconover: I haven't been run your tests yet but I will get to them today :wink:

simov: Ok, just don't forget to execute the commands from my first comment https://github.com/request/request/pull/1966#issuecomment-164790235 you may not need all of them though.

Also I just added the:

```bash
# obtain a ticket-granting ticket using the keytab
$ kinit -k -t /etc/krb5.keytab HTTP/localhost
```

Not sure how is this supposed to work in the real world, I mean I can't find such method in the `kerberos` module.

rustyconover: Don't worry about this test yet.  I think it may be considered strange to use localhost when using Kerberos credentials but I'll figure this all out. As services are bound to hostnames upon registration if you if you have a KDC that is allowing binds to HTTP/localhost and you have multiple machines running services on localhost you're not going to be able to tell which localhost you're referring to.

Hence you may need to bind the service to an actual hostname.

I think I know enough Kerberos to try to make the test work, and thankfully I know other people who know way *more than enough* to help me out with any problems, @sdegler that's you.

simov: That makes sense. I'm using `localhost` just for the test, because the test server is listening on `localhost:6767`. The tests are actually working on my end, so you should be able to run them if you have everything setup. Of course I'm open for suggestions and improvements :+1:

rustyconover: I installed your setup on a new Docker ubuntu image.  I think you've got most of the way there.  What did you use as the default realm name, I just picked to use `EXAMPLE.COM`

Added an entry for `kdc.example.com` to point to localhost in /etc/hosts.

Then ran:

 ```
# I'm running as root in the container, so I don't have sudo before every command.
# This will set a password for the realm.
echo "test123" | krb5_newrealm

# Add a root principal, again with a silly password
kadmin.local -q "addprinc -pw test123 root@EXAMPLE.COM"

# Authenticate into the realm, this will establish a TGT
echo "test123" | kinit root@EXAMPLE.COM

# Create the service principal for the HTTP service, localhost is okay after all.  It will implicitly have the default realm name appended on to it.  So the real principal will be HTTP/localhost@EXAMPLE.COM
kadmin.local -q "addprinc -randkey HTTP/localhost"

# Add the service principal to root's keytab.
kadmin.local -q "ktadd HTTP/localhost"

# Then run the tests and everything passes.
npm run test
```

So yes the tests work for me too, great job! :tada:

**Comment by rustyconover** [comment]
@simov ping?

**Comment by simov** [comment]
Yep time is passes by .. I was trying to automate the kerberos installation and setup on docker, then I was planning to reproduce the same steps on TravisCI. So even if you are working on request on some other OS than Linux locally, you can still rely on TravisCI at least.

The problem is that the tests that I added works only for two of us, and I'm not going to add your implementation without tests. I can't even remember where exactly I was stuck, but essentially that's the problem - we need to automate the steps outlined by two of us and run the kerberos tests on TravisCI at least.

**Comment by rustyconover** [comment]
I was talking with my friend Roland about this predicament of needing to
setup a KDC, there may be a lighter weight alternative.  I'll get more
concrete details from him and share them.

Rusty

Sent from my iPhone

On Dec 29, 2015, at 4:30 PM, simo <notifications@github.com> wrote:

Yep time is passes by .. I was trying to automate the kerberos installation
and setup on docker, then I was planning to reproduce the same steps on
TravisCI. So even if you are working on request on some other OS than Linux
locally, you can still rely on TravisCI at least.

The problem is that the tests that I added works only for two of us, and
I'm not going to add your implementation without tests. I can't even
remember where exactly I was stuck, but essentially that's the problem - we
need to automate the steps outlined by two of us and run the kerberos tests
on TravisCI at least.

—
Reply to this email directly or view it on GitHub
<https://github.com/request/request/pull/1966#issuecomment-167881758>.

**Comment by rustyconover** [comment]
So I've heard back, the trick is to use:

kimpersonate

http://manpages.ubuntu.com/manpages/natty/man8/kimpersonate.8.html

It is packaged in the Heimdal package.

**Comment by cbayram** [comment]
Thanks for this work, this saved me time on integration. I look forward to it being merged to master.

Fyi, 302 redirects are failing currently against kerberized HTTP services.  I'm investigating this issue and will post my findings.

File: lib/auth.js [hunk 10]
```diff
@@ -48,6 +49,60 @@ Auth.prototype.bearer = function (bearer, sendImmediately) {
...
+  if (!sendImmediately) {
+    return
+  }
+
+  var k = new kerberos.Kerberos()
+  var service_identifier = 'HTTP@' + request.host
+
+  self.pendingAuth = true
+  k.authGSSClientInit(service_identifier,
+    kerberos.Kerberos.GSS_C_MUTUAL_FLAG,
```
cbayram: ```kerberos.Kerberos.GSS_C_MUTUAL_FLAG | kerberos.Kerberos.GSS_C_DELEG_FLAG,```
will support multi-hop delegations of the kerberos credentials.

request.js
```diff
@@ -390,12 +390,16 @@ Request.prototype.init = function (options) {
     if (Object.prototype.hasOwnProperty.call(options.auth, 'password')) {
       options.auth.pass = options.auth.password
     }
+    if (Object.prototype.hasOwnProperty.call(options.auth, 'negotiate')) {
+      options.auth.negotiate = options.auth.negotiate
+    }
```
cbayram: this is unnecessary/redundant.

cbayram: This is unnecessary/redundant.

request.js
```diff
@@ -390,12 +390,16 @@ Request.prototype.init = function (options) {
     if (Object.prototype.hasOwnProperty.call(options.auth, 'password')) {
       options.auth.pass = options.auth.password
     }
+    if (Object.prototype.hasOwnProperty.call(options.auth, 'negotiate')) {
```
cbayram: On 302 redirects, we never get back into the ```if (options.auth) {``` block to re-negotiate on the redirected path.

stillesjo:
@rustyconover @simov 
Would it be possible to continue working on this? 
Is it still tests that needs to be created for this? Anyway I can help?

rustyconover:
Sure, I'll revisit it shortly.

rmanyari:
Rodrigo from Two Sigma here.

The last commit has a few things that hopefully will get us closer to merge this to master:

- Dockerfile with KDC
- Integration tests for kerberos
- kerberos is declared as an optional dependency. IMO this is very important since we don't want to impose it to everyone, specially since it requires the C++ headers...

Although everything hangs together nicely, there are a few things that we could improve:

- Push the docker image to the public docker registry, ideally under a 'request' account. That way we don't have the build the image every single time.
- I excluded kerberos from the Browserify process. I'm not too familiar with it. If anyone is brave enough, we can add it back.
- npm test now triggers npm run test-local (former npm test) and npm run test-integration (test suite that assumes a travis-ci like environment). I added this npm run test-local for convenience, not every contributor might want to install docker and libkrb5... In they want to do so, the steps are documented in CONTRIBUTING.md

@simov - lmk what you think, hopefully it's in the right direction

Signed-off-by: Rodrigo Manyari rodrigo@twosigma.com

FredKSchott:
I'm +1 on supporting some Kerberos solution for request (supporting many different situations and scenarios is what makes request so useful). However, bundling a somewhat-niche feature like Negotiate/SPNEGO/Kerberos support into the request library itself is not the best way to offer this support, in my opinion. Here are the drawbacks I'm seeing:
  - Our test infrastructure is made much more complex with the introduction of docker and a configured kerberos server.
  - Optional dependencies are poorly supported in npm and yarn, and commonly cause unexpected version mismatches and warnings on install time.
  - the `auth()` function(s) are now further overloaded, where `negotiate`/`user & pass`/`bearer` all override and conflict with each other .
  - As far as I know, no contributors are familiar enough with Kerberos to reasonably fix any bugs that may come up with owning the implementation.

But the actual kerberos auth negotiation logic is actually pretty self contained (as written in `Auth.prototype.negotiate()`).  Instead of jamming it around the current Auth type, what if request could take an optional `auth`  option for handling complex/custom auth schemes? That way request could support kerberos and others without owning each and every implementation. This is much much closer to the support we imagined when we first pulled auth logic off of the request object itself.

Here's a strawman of what I'm thinking:

```js
const KerberosAuth = require('request-kerberos');
request({
  url: 'example.com',
  more: {'options': 'here'},
  auth: KerberosAuth,
});
```

```js
function KerberosAuth(request) {
  // define all public properties here
  this.request = request
  this.hasAuth = false
  this.sentAuth = false
}

KerberosAuth.prototype.onRequest = function (options) {
  self.negotiate(this.request, undefined, options.sendImmediately);
}

KerberosAuth.prototype.onResponse = function (response) {
  if (!self.hasAuth || self.sentAuth) { return null }
  var authHeader = caseless(response.headers).get('www-authenticate')
  var authVerb = authHeader && authHeader.split(/,?\s+/)[0].toLowerCase()
  if (authVerb === 'negotiate') {
    self.negotiate(request, authHeader, true);
  }
}

KerberosAuth.prototype.negotiate = function(request, authHeader, sendImmediately) {
  // see your implementation in #1966
}
```

@rustyconover Would that accomplish your goals here? What do you think?

PR #1984, opened 2015-12-29
Move timeout code to after a socket is assigned

evanderkoogh:
In the old code the timeout was started when the request was queued, not when the
request was send to the server. If maxSocket option is set on an Agent,
the actual request might be send later. This is not what is suggested in the documentation.
Code to reproduce the problem can be found in this gist: https://gist.github.com/evanderkoogh/e7e94f036c7a7a6b8f29

I have moved all of the code related to setting up timeouts to self.req.on('socket'). This event is called when the request has been assigned a socket. After changing my local copy the code above runs perfectly now.

### DIFF
--- a/request.js
+++ b/request.js
@@ -749,41 +749,6 @@ Request.prototype.start = function () {
     self.startTime = new Date().getTime()
   }
 
-  if (self.timeout && !self.timeoutTimer) {
-    var timeout = self.timeout < 0 ? 0 : self.timeout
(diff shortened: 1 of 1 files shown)


evanderkoogh:
Is there any way to rerun that Travis check? It has a timeout on a test that reads from disk. I haven't changed anything related to that and all other environments are fine. Most likely an issue with the environment instead of with the code.

PR #2009, opened 2016-01-13
Support JSON stringify replacer argument.

elyobo:
Adds support for the replacer argument of JSON.stringify, in the same
way that request already supports the revivier argument for JSON.parse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -749,6 +749,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 - `postambleCRLF` - append a newline/CRLF at the end of the boundary of your `multipart/form-data` request.
 - `json` - sets `body` to JSON representation of value and adds `Content-type: application/json` header.  Additionally, parses the response body as JSON.
 - `jsonReviver` - a [reviver function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) that will be passed to `JSON.parse()` when parsing a JSON response body.
+- `jsonReplacer` - a [replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) that will be passed to `JSON.stringify()` when stringifying a JSON request body.
 
 ---
 
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -24,12 +24,12 @@ function paramsHaveRequestBody(params) {
   )
 }
 
-function safeStringify (obj) {
+function safeStringify (obj, replacer) {
   var ret
   try {
-    ret = JSON.stringify(obj)
+    ret = JSON.stringify(obj, replacer)
   } catch (e) {
-    ret = jsonSafeStringify(obj)
+    ret = jsonSafeStringify(obj, replacer)
   }
   return ret
 }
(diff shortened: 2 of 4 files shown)


elyobo:
Updated to remove extra semicolon.

elyobo:
@mikeal Anything I can do to move this one forward?  It seems like a simple change and complements the existing use of the reviver when parsing JSON, but I haven't heard a yeah, a nay or a please-change-this from anyone.

mikeal:
So, I'm +1 on adding this *for now* but it should be named something else, like "jsonStringify".

elyobo:
Thanks @mikeal.

Just to explain my reasoning, I went with `jsonReplacer` as, like the existing `jsonReviver` argument, it reflects the name of the argument parsed to the underlying [JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) call (as reviver reflects the argument to [JSON.parse](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse)) and I thought that those names would be more obvious to users because of that.

That said, if you're sure that something like `jsonStringifier` instead I'll make that change and repush.

elyobo:
Possibly something like `jsonParseReviver` and `jsonStringifyReplacer` would be more descriptive for both of them.

elyobo:
Test failures appear unrelated...

README.md
```diff
@@ -749,6 +749,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 - `postambleCRLF` - append a newline/CRLF at the end of the boundary of your `multipart/form-data` request.
 - `json` - sets `body` to JSON representation of value and adds `Content-type: application/json` header.  Additionally, parses the response body as JSON.
 - `jsonReviver` - a [reviver function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) that will be passed to `JSON.parse()` when parsing a JSON response body.
+- `jsonReplacer` - a [replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) that will be passed to `JSON.string()` when stringifying a JSON request body.
```
simov: The function is called `JSON.stringify()` not `JSON.string()`

elyobo: Ta, fixed.  Got it right in the URL, not sure what I was thinking in the comment.

simov:
@elyobo don't squash commits before the final review, now I can't see what your first commit looked like.

`jsonReplacer` is fine, that's how we add those options so far. Ideally we should be using an object for these settings like [this](https://github.com/request/core#parse)

elyobo:
Apologies on the squashing, I can probably reconstruct it out of the reflog if you'd like to see it how it was.  The only change in the last one was to correct the `JSON.string()` to `JSON.stringify()` in the readme.  Before that the change was to update the docs in the readme, which I'd missed before that.  Before that was a change to switch to jsonStringifier along the lines of mikeal's suggestion which I pushed accidentally, then reverted and squashed.  No functional changes have been made at any time.

simov:
Ok :+1: do you want this change published soon? It's probably a good time for new release anyway.

elyobo:
Thanks for the merge, whenever suits you to release is fine by me.

PR #2018, opened 2016-01-19
Fix remote memory disclosure with multipart attachments

feross:
If the node process makes a request with a multipart attachment, and the type of the `body` option is a `Number`, then that many bytes of uninitialized memory will be sent in the body of the request. Here's an example:

```js
request({
  method: 'POST',
  uri: 'http://localhost:8000',
  multipart: [{ body: 1000 }]
}
```

And here's a more complete reproducible example:

```js
var http = require('http')

http.createServer(function (req, res) {
  var data = ''
  req.setEncoding('utf8')
  req.on('data', function (chunk) {
    console.log('data')
    data += chunk
  })
  req.on('end', function () {
    // this will print uninitialized memory from the client
    console.log('Client sent:\n', data)
  })
  res.end()
}).listen(8000)

request({
  method: 'POST',
  uri: 'http://localhost:8000',
  multipart: [{ body: 1000 }]
},
function (err, res, body) {
  if (err) return console.error('upload failed:', err)
  console.log('sent')
})
```

This PR fixes the issue by coercing the type of the `body` option to a string. (An alternate solution would be to throw an exception in this case.)

Further reading:

- https://github.com/nodejs/node/issues/4660
- https://github.com/nodejs/node-eps/pull/4

### DIFF
--- a/lib/multipart.js
+++ b/lib/multipart.js
@@ -68,6 +68,9 @@ Multipart.prototype.build = function (parts, chunked) {
   var body = chunked ? new CombinedStream() : []
 
   function add (part) {
+    if (typeof part === 'number') {
+      part = part.toString()
+    }
     return chunked ? body.append(part) : body.push(new Buffer(part))
   }
 


feross:
For the record, I'm sharing this here for the first time since this doesn't seem particularly easy to exploit. That said, it's still probably a good idea to release a fix in a timely manner.

simov:
Thanks I've added your commit here https://github.com/request/request/pull/2022 + a test case.

ChALkeR:
@feross 
> For the record, I'm sharing this here for the first time since this doesn't seem particularly easy to exploit. That said, it's still probably a good idea to release a fix in a timely manner.

It still looks like a much better idea to send those over emails. Can I have a chat with you on IRC or Gitter?

ChALkeR:
@feross I know of at least one module that reads config from a JSON file and sends a multipart request using that value as the raw `body` entry of some multipart data to a remote server.

grnd:
Vulnerable range: `<2.68.0 >2.2.5` (total of 77 releases)
There is one version in that range that's not exploitable, `2.47.0` (throws TypeError).
While`<=2.2.5` is not vulnerable because part's body is wrapped with '\r\n' just before it is passed to Buffer.
```javascript
      self.body += '\r\n' + body + '\r\n'
    })
    self.body += '--frontier--'
  }

  if (self.body) {
    if (!Buffer.isBuffer(self.body)) {
      self.body = new Buffer(self.body)
    }
```

evilaliv3:
I just found out that relevant and widespread softwares are still impacted:
- https://github.com/nodejs/node-gyp/pull/1000
- https://github.com/mapbox/node-pre-gyp/pull/232
- https://github.com/yeoman/insight/pull/48
- https://github.com/FGRibreau/node-request-retry/pull/37
- https://github.com/angular/angular.js/issues/14961
- https://github.com/angular/angular/issues/10352
- https://github.com/strongloop/fsevents/issues/145
- https://github.com/axemclion/grunt-saucelabs/issues/214

I think that it should be important to inform better the users around issues like this.

ChALkeR:
@evilaliv3 There are about 14000 of those.

ChALkeR:
Ah, sorry, forgot the lower limit. About 10400.

PR #2021, opened 2016-01-20
Fix security vulnerabilities

sdepold:
har-validator uses is-my-json-valid: https://nodesecurity.io/advisories/76
hawk: https://nodesecurity.io/advisories/77

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,13 +35,13 @@
     "tough-cookie": "~2.2.0",
     "http-signature": "~1.1.0",
     "oauth-sign": "~0.8.0",
-    "hawk": "~3.1.0",
+    "hawk": "~4.1.1",
     "aws-sign2": "~0.6.0",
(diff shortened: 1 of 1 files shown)


sdepold:
🙄Hm. Apparently hawk is no longer compatible with node <4.

simov:
Yep :) you can upgrade your hawk version manually (in case you use it with request). Is there any other problem with not upgrading this dependency that I'm no aware of?

sdepold:
Hm OK. I'm not sure how this would help though since hawk would still install the "broken" version of hawk.

simov:
Yep, you are correct. I still think that migrating request to Node > 4.0.0 will cause much larger disruption than hawk being with security vulnerability. It's just that currently request have too many hard dependencies that are not essential to its core functionality.

sdepold:
I agree.

PR #2024, opened 2016-01-20
Update har-validator dependency for nsp advisory #76

TylerDixon:
See https://nodesecurity.io/advisories/76


simov:
Thanks, I'm just letting you know that our version range installs the latest patch version anyway.

PR #2026, opened 2016-01-21
update hawk for security fix, see https://nodesecurity.io/advisories/77

ralphtheninja:
Fixes

request-nsp


simov:
Read here https://github.com/request/request/pull/1891#issuecomment-173671203

PR #2036, opened 2016-01-26
Add AWS Signature Version 4

simov:
Closes https://github.com/request/request/pull/2031

- [AWS Signature Version 4](http://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
https://www.npmjs.com/package/aws4)
- [aws4](https://www.npmjs.com/package/aws4) module

PR #2041, opened 2016-01-27
restore aws4 as regular dependency

### LINKED ISSUE #2040: Breaking Change in v2.68.0 w/ Browserify
Looks like v2.68.0 introduced a breaking change by requiring a separate `aws4` dependency.  Apparently with browserify, that dependency isn't optional, causing build failures.

rmg:
It was moved in #2036 as a devDependency, but is still used in
request.js

This should fix #2040.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "main": "index.js",
   "dependencies": {
     "aws-sign2": "~0.6.0",
+    "aws4": "^1.2.1",
     "bl": "~1.0.0",
     "caseless": "~0.11.0",
     "combined-stream": "~1.0.5",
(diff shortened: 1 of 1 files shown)


mikeal:
released v2.69.0.

rmg:
@mikeal thanks, verified. :+1:

simov:
Oh, wow, guess who's not using browserify .. also used during runtime and detected during parsing the file as a string are two completely different things. Point taken - never trust browserify :) 

Also I was thinking about adding it as peerDep but I wasn't sure about the warning, otherwise not installing it (if npm3 is used) is what I want. As for the optionalDep that's completely useless in my case because the dep is still being installed.

mikeal:
we should never use peerDeps, they are a giant mistake :)

simov:
Well installing them automatically is a mistake I agree. Warnings shouldn't break the build though.

PR #2052, opened 2016-02-03
Cancel timeout on aborted requests

### LINKED ISSUE #2025: abort() not working
```JavaScript
var ro = request({url: 'http://google.com:81', timeout: 20000}, function(e, r, b) {console.log('done');});
ro.abort();
```

20 seconds later, I still get the "done" output. Expected behaviour: either don't call the callback at all, or call the callback right away with one of the parameter having some kind of `wasAborted` flag.

stephenplusplus:
Fixes #2025

When a request with a `timeout` is aborted, the setTimeout needs to be cleared as well. Please let me know if I didn't think of anything, or otherwise didn't make a satisfactory PR!

### DIFF
--- a/request.js
+++ b/request.js
@@ -1041,6 +1041,11 @@ Request.prototype.abort = function () {
   var self = this
   self._aborted = true
 
+  if (self.timeout && self.timeoutTimer) {
+    clearTimeout(self.timeoutTimer)
+    self.timeoutTimer = null
+  }
+
   if (self.req) {
     self.req.abort()
   }
@@ -1229,10 +1234,10 @@ Request.prototype.aws = function (opts, now) {
     self._aws = opts
     return self
   }
-  
+
   if (opts.sign_version == 4 || opts.sign_version == '4') {
     var aws4 = require('aws4')
-    // use aws4  
+    // use aws4
     var options = {
       host: self.uri.host,
       path: self.uri.path,
--- a/tests/test-timeout.js
+++ b/tests/test-timeout.js
@@ -138,6 +138,19 @@ if (process.env.TRAVIS === 'true') {
     })
   })
 
+  tape('abort', function (t) {
+    var shouldTimeout = {
+      url: s.url + '/timeout',
+      timeout: 10000
+    }
+
+    var req = request(shouldTimeout)
+    // This timer should be cleared:
+    req.timeoutTimer = setTimeout(t.end, 100)
+    req.abort()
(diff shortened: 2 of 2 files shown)


request.js
```diff
@@ -1229,10 +1234,10 @@ Request.prototype.aws = function (opts, now) {
     self._aws = opts
     return self
   }
-  
+
   if (opts.sign_version == 4 || opts.sign_version == '4') {
     var aws4 = require('aws4')
-    // use aws4
```
partycoder: this sucks

partycoder:
This doesn't call the callback it seems?

request.js
stephenplusplus: White space removal? Happened automatically from using my editor. The whitespace was obviously an accident to begin with. If the maintainers want it removed, I'm happy to.

stephenplusplus:
Does abort get a callback?

partycoder:
How else would you from an application standpoint handle errors?
e.g: retry, log, send a response, etc?

stephenplusplus:
I'm just looking at the code: https://github.com/request/request/blob/7f7b2783ccecc7af79c0f74faeb0d34a3ac37d56/request.js#L1040

partycoder:
You would need to listen to the 'abort' event, which would be a really inconsistent way of handling your error.

stephenplusplus:
> Today your callback gets called with an error ('ETIMEDOUT'), removing this will basically not generate anything at all.

That's because of the bug this PR is meant to fix. If you set a timeout, then abort the request before the timeout clears (say after 10 seconds while your timeout was set to 30), the `setTimeout` isn't cleared, so it'll be another 20 seconds before your callback is triggered with the `ETIMEDOUT` error that was created by this library.

I can't speak for the authors, but I think of aborting a request as you abruptly destroying the connection for your own reasons. Like, "the user clicked 'Cancel'-- I don't care about this GET anymore". There is no error to handle, because you are the one doing the destroying.

partycoder:
OK, so basically in the case the request is aborted the callback won't be called. This needs to be documented then.

The other thing is: abort affects piped requests too. How would events propagate in this case?

case 1. `request.pipe(res)`
case 2. `req.pipe(request).pipe(res)`

stephenplusplus:
The stream is destroyed, which allows for internal communication between streams that are connected. Play around with request as it is now, just don't set a timeout on your requests.

PR #2057, opened 2016-02-04
Update contributing guidelines

simov:
@request/contributors I'm going to introduce two handy labels:

- `Help (please use Stackoverflow)`
- `Not enough info (see CONTRIBUTING.md)`

So that it will be easier to filter out the noise + closing issues without further notice (the first case).


eiriksm:
:+1:

PR #2058, opened 2016-02-04
Karma fixes for latest versions

### LINKED ISSUE #2056: Karma tests fail
@eiriksm both https://github.com/request/request/pull/2035 and https://github.com/request/request/pull/2044 are failing due to some weird error can you take a look at it? Just install those versions and you'll see the error. Thanks!

eiriksm:
So, here is some fixes that closes #2035 and #2044 and #2056

First off, one of the issues were that we are testing across 3 versions of npm. They have different behaviours when encountering peerDependencies. So ideally I would almost say we should upgrade npm to at least npm@2 for node 0.10. But open to opinions there. This was the least disruptive change, in my opinion.

Second, one of the issues were pointing out that the response object changed for one of the tests. So I ran the same test code with node, where the result was illuminating the fact that the browser test was testing for something only happening in the browser. So I changed that test, so it can check for a result that also is the same result when running on node.

Fixes #2035
Fixes #2044
Closes #2056


eiriksm:
Also, let me know if I should squash these commits. Had to do a couple of runs to get the suite to run it on travis to be sure :)

simov:
Your commits are great for explaining what's going on. I often find myself scrolling through the last commits for a project, so from a usability point of view clicking on a single and more meaningful commit there is easier. Of course this can be approach in many different ways, but in this case one commit might be better.

Also see how I'm using the `Fixes` and `Closes` to actually close those PRs and Issues once this gets merged in. Probably another syntax might work too, but I don't think GitHub is smart enough to link those issues from:

> So, here is some fixes that closes #2035 and #2044 and #2056

Thanks for taking the time to work on this :+1:

eiriksm:
>Also see how I'm using the Fixes and Closes to actually close those PRs and Issues once this gets merged in

I am aware of the syntax :) Just wanted to make sure the issues were being referenced, for easier navigation.

Updated the branch and used this to indicate what is being closed and fixed.

simov:
Thanks :cat2:

PR #2066, opened 2016-02-10
Update vulnerabilities in deps badge

remy:
The badge that was being used was *incorrectly* showing that request held vulns in it's (production) dependencies, but this is wrong. 

Request uses hawk, which _had_ a vuln, but has been patched and released in 3.1.3 (and is [vuln free](https://snyk.io/test/npm/hawk/3) in the 3.x branch), and the semver range for hawk in the `package.json` is `~3.1.0` - so 3.1.3 is going to be installed.

Disclosure: I've been working on snyk since mid-last year, but I've been relying on request for even longer and I didn't want it be wrongly represented as vulnerable 💙


simov:
Having snyk badge there would be great! I'm just wondering whether setting up hawk to 3.1.3 would make david-dm happy. I don't mind having both badges there, I think david-dm is strictly for showing out of date dependencies.

PR #2079, opened 2016-02-15
Accept read stream as body option

### LINKED ISSUE #2075: Accept streams as body
If I want to send the request body e.g. gzipped, I currently need the return value and pipe the gzipstream there. However I only have access to the return value when using callback API but we are normally using promise API where the return value "slot" is taken by the promise and we cannot interact with the stream.

I propose extending the `body` option to accept stream which would do this for me so that I don't need to waste the return value "slot". Any errors should be redirected to the callback error argument.

simov:
Fixes #2075

### DIFF
--- a/README.md
+++ b/README.md
@@ -733,7 +733,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 
 ---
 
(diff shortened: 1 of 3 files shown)


petkaantonov:
lgtm

simov:
Let me know if you need this published on NPM

PR #2088, opened 2016-02-21
Replace lib/getProxyFromURI.js with proxy-from-env

Rob--W:
I was looking for a package that implements support for the (de facto) standard proxy environment variables, but didn't find any except for getProxyFromURI.js in this library. My two main issues with the implementation is that it uses suffix matching (e.g. `NO_PROXY=example.com` also matches `notexample.com`) and that it uses `HTTP_PROXY` even for https URLs). The latter is problematic if I want to proxy all HTTP requests, but not HTTPS requests, because there is no way to get `NO_PROXY` to stop matching https requests.
I've fixed all of the above issues and published the result as https://www.npmjs.com/package/proxy-from-env, and decided to submit a PR here so that eventually all Node.js applications/libraries that want to support setting the proxy via the environment variables do so in a consistent and standard way.

Compared to the existing implementation in request, there are two backwards-incompatible changes (see the commit message for other differences and more details):
- `HTTP_PROXY` is no longer used for https requests (use `HTTPS_PROXY`
  or `ALL_PROXY`).
- The items in `NO_PROXY` are no longer used in suffix matching, but used as an exact match (unless the item starts with a dot or asterisk, then suffix matching will be used).

The proxy-from-env library [has an extensive set of tests](https://github.com/Rob--W/proxy-from-env/blob/master/test.js). With a small modification, the tests can also be used to check the existing implementation:

``` sh
git clone https://github.com/Rob--W/proxy-from-env.git
cd proxy-from-env

# Get current implementation from request
wget https://github.com/request/request/blob/c2dab797d362ad158e0399140aa59829fa50d9b3/lib/getProxyFromURI.js?raw=true -Oindex.js

# Change module signature to match the library.
echo "module.exports={getProxyForUrl: function(s) { return getProxyFromURI(require('url').parse(s)) || ''; }};" >> index.js

# Install dependencies and run tests
npm install
npm test
```

Here is the test output: http://pastebin.com/raw/Dk9f79Vq (anything without a checkmark is a test failure).


simov:
That's great, but I'm worried about breaking changes as we don't follow semver for the current major version 2.x. At this point a better approach would be to pass an option to request and specify which `getProxyFromURI` implementation you want to use.

Rob--W:
What do you mean by "I'm worried about breaking changes **as we don't follow semver for the current major version 2.x.** "?

The [README of request states](https://github.com/request/request#controlling-proxy-behaviour-using-environment-variables):

> When `HTTP_PROXY` / `http_proxy` are set, they will be used to proxy non-SSL requests that do not have an explicit proxy configuration option present.

The "non-SSL" qualifier rules out that the variable can be used for HTTPs requests (even though the implementation falls back to `http_proxy` when `https_proxy` is missing). And I've checked many implementations of the proxy, and none of them use `http_proxy` for proxing HTTPS requests (see the README of the proxy-from-env library for sources that I checked), so users cannot realistically expect that `http_proxy` is also used for https URLs.
Assuming that users don't rely on undocumented features, merging this PR should be safe.

simov:
Ok, I'm getting your point now, let's hope for the best! 

cc @jvmccarthy @samcday @seanstrom

Rob--W:
I rebased the PR to resolve the merge conflict in package.json. Did you already have a chance to review it?

FredKSchott:

Rob--W:
@FredKSchott I rebased my branch on master and verified locally that the tests still pass. The PR is not automatically updated because you've closed it though.

FredKSchott:
@Rob--W Oh damn and I can't re-open this one either... mind creating a new PR with a link to this one?

PR #2100, opened 2016-02-26
Proposal to allow setting of timeout with env var

cpb:
When I start work with an unfamiliar open source project I try to make sure the tests pass. Some of these projects use request. Sometimes, the connection or requests their tests make timeout. It would be nice if I could get the tests to pass without modifying any of the code of a build that passes in other contexts. Being able to set a different timeout across all the calls made with request is one way that I can work around exceptional latency, without changing the requirements of the open source project under test.

I am curious as to your own approach for debugging this. Sometimes unfamiliar open source projects layer abstractions on top of request. It can be difficult to find where to provide an alternate timeout value.

### DIFF
--- a/request.js
+++ b/request.js
@@ -128,6 +128,10 @@ function Request (options) {
   extend(self, nonReserved)
   options = filterOutReservedFunctions(reserved, options)
 
+  if (parseFloat(process.env.REQUEST_TIMEOUT) > 0) {
+    self.timeout = parseFloat(process.env.REQUEST_TIMEOUT)
+  }
+
   self.readable = true
   self.writable = true
   if (options.method) {
--- a/tests/test-timeout.js
+++ b/tests/test-timeout.js
@@ -98,6 +98,52 @@ if (process.env.TRAVIS === 'true') {
     })
   })
 
+  tape('generous process.env.REQUEST_TIMEOUT', function(t) {
+    process.env.REQUEST_TIMEOUT = 2000
+
+    var noTimeoutInOptions = {
+      url: s.url + '/timeout'
+    }
+
+    request(noTimeoutInOptions, function(err, res, body) {
+      t.equal(err, null)
+      t.equal(body, 'waited')
+      t.end()
+      process.env.REQUEST_TIMEOUT = null
+    })
+  })
+
+  tape('ambitious process.env.REQUEST_TIMEOUT', function(t) {
+    process.env.REQUEST_TIMEOUT = 100
+
+    var noTimeoutInOptions = {
+      url: s.url + '/timeout'
+    }
+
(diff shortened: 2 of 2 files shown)


mikeal:
Is there a reason this needs to be global?

Would this not handle most cases:

```javascript
var request = require('request').defaults({timeout:process.env.REQUEST_TIMEOUT})
```

cpb:
Thanks for this insight!

cpb:
This may be good enough for my use case, however, this isn't quite the behaviour of the pull request. In the pull request the behaviour is an override rather than a default value.

simov:
@cpb I though you want to set a timeout over a third party test suite, one that you can't modify. So is the `defaults` approach working for you?

cpb:
Not really. They specify their own timeout, which is still too low under my
network conditions. Maybe having an ```overrides``` convenience method is
what I am looking for. I thought the ```REQUEST_TIMEOUT``` env variable was
similar to the one for setting a proxy.

TBH it does feel a bit hacky for this specific use case. The question is,
do you want to support this for other as yet unimagined use cases?

simov:
Options passed to request should override global env vars, I think. Not vice versa.

cpb:
I am glad we can have this discussion. It was slightly counter to my
intuition for how this gloval env var might behave, hence I tried providing
tests to support that clarification! In some http libraries, I believe the
proxy env var _is_ an override. But I get the sense that this
env-var-as-override feels a bit kludgy.

simov:
It depends on how you look at it, whether you want all request in that process to be stuck with certain timeout, or whether you want certain requests to be able to override the process wide timeout.

BTW you can set the timeout after the request have been initialized (before being started):

```js
var http = require('http')
var server = http.createServer()
server.on('request', (req, res) => {
  setTimeout(() => {
    res.end('end')
  }, 3000)
})
server.listen(3000, () => {
  var r = request('http://localhost:3000', (err, res, body) => {
    if (err) {
      console.log(err)
    }
    else {
      console.log(body)
    }
  })
  r.timeout = 2000
})
```

No idea if that works for you.

cpb:
Wow cool!

mikeal:
I'm hesitant to consider any globals like this for a few reasons:

* Different versions of request will set these module level globals separately *in the same application process*, which means that any libraries requiring older versions before this feature was introduced will have severely differing behavior.
* Request tries desperately to localize configuration and state, even going as far as creating our own pooling API to work around the global pool in Node.js.
* Request is entirely built on the underlying Node.js http client which actually *is* process global and relies on this fact for several features (like pooling). If there's a place for a global timeout hack it should probably go in the http client rather than request.

cpb:
Thanks Mikeal, I really appreciate the time you and your team have put to
this question!

PR #2113, opened 2016-03-04
Extra optional parameter

ShaharHD:
Allows the user to define using the options an extra option which he can use inside the callback functions.

This is important when you use request inside any loop, as it is not recommended to define callback functions inside a loop but rather use a pre-defined method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Request is designed to be the simplest way possible to make http calls. It suppo
 
 ```js
 var request = require('request');
-request('http://www.google.com', function (error, response, body) {
+request('http://www.google.com', function (error, response, body, extra) {
   if (!error && response.statusCode == 200) {
     console.log(body) // Show the HTML for the Google homepage.
   }
@@ -240,7 +240,7 @@ Some variations in different HTTP implementations require a newline/CRLF before,
       ]
     }
   },
-  function (error, response, body) {
+  function (error, response, body, extra) {
     if (error) {
       return console.error('upload failed:', error);
     }
@@ -301,7 +301,7 @@ var username = 'username',
     password = 'password',
     url = 'http://' + username + ':' + password + '@some.server.com';
 
-request({url: url}, function (error, response, body) {
+request({url: url}, function (error, response, body, extra) {
    // Do more stuff with 'body' here
 });
 ```
@@ -339,7 +339,7 @@ var options = {
   }
 };
 
-function callback(error, response, body) {
+function callback(error, response, body, extra) {
   if (!error && response.statusCode == 200) {
     var info = JSON.parse(body);
     console.log(info.stargazers_count + " Stars");
@@ -721,6 +721,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 - `baseUrl` - fully qualified uri string used as the base url. Most useful with `request.defaults`, for example when you want to do many requests to the same domain.  If `baseUrl` is `https://example.com/api/`, then requesting `/end/point?test=true` will fetch `https://example.com/api/end/point?test=true`. When `baseUrl` is given, `uri` must also be a string.
 - `method` - http method (default: `"GET"`)
 - `headers` - http headers (default: `{}`)
+- `extra` - extra data which will be passed to the callback (default: `undefined`)
 
 ---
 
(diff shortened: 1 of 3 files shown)


simov:
Not sure what your goal exactly is, but ideally this logic should be handled in your application.

ShaharHD:
The goal was an easier way to transfer data and/or parameters to a predefine callback method (which is shared between multiple requests) or requests being done from a loop (and limited using the pool variable)

Here is a code snippet which might shed light on where I use it:

```
for (var portIndex in ports) {

    var port = {
        name: ports[portIndex].name
        ...
    };

    var address = {
        city: ports[portIndex].city
    };

    var location = {
        latitude: ports[portIndex].latitude,
        longitude: ports[portIndex].longitude,
        altitude: ports[portIndex].altitude
    };

    request.put({
        url: api,
        pool: requestPool,
        json: true,
        extra: {
            address: address,
            location: location
        },
        body: port
    }, processResponse);

}


function processResponse(err, httpResponse, body, extra) {
    if (err !== null) {
        console.log(err);
    } else {
        if (body.id !== undefined) {
            // We have an id, let's write address and location

            if (extra.address !== undefined) {
                request.put({
                    url: api + '/' + body.id + '/address',
                    pool: requestPool,
                    json: true,
                    body: extra.address
                }, processAddress);
            }

            if (extra.location !== undefined) {
                request.put({
                    url: api + '/' + body.id + '/location',
                    pool: requestPool,
                    json: true,
                    body: extra.location
                }, processLocation);
            }
        }
    }

}

function processAddress(err, httpResponse, body) {
  console.log(JSON.stringify(body));
}

function processLocation(err, httpResponse, body) {
  console.log(JSON.stringify(body));
}
```

simov:
How about that:

```js
var obj = {a: 1}
function processResponse (extra, err, res, body) {
}
request.get('http://google.com', processResponse.bind(null, obj))
```

ShaharHD:
As I understand it, it created a new method for every for loop call, which is almost as doing the not so recommended:

```
for (var portIndex in ports) {

    var port = {
        ...
    };

    var address = {
        ...
    };

    var location = {
        ...
    };

    request.put({
        url: api,
        json: true,
        body: port
        //can access address & location here
    });

}
```

simov:
```js
var obj = {a: 1}

var processResponse = function (extra, err, res, body) {
}.bind(null, obj)

for (var i=0; i < 10000000000000000; i++) {
  request.get('http://google.com', processResponse)
}```

**Comment by ShaharHD** [comment]
Take a look at my original code snippet loop - the data changes between the loop iterations.
If it was a static data - like your for examples - then bind will work just fine.

But I might be totally wrong here?

**Comment by simov** [comment]
With the current implementation this works too:

```js
var obj = {a: 1}
request.get('http://google.com', {extra: obj},
function (err, res, body) {
  console.log(this.extra)
})
```

simov:
In case you don't want to rely on `this` you can simply use a closure:

```js
function processResponse (extra) {
  return function (err, res, body) {
    console.log(extra)
  }
}

for (var i=0; i < 3; i++) {
  var obj = {a: i}
  request.get('http://google.com', processResponse(obj))
}
```

Closures are fast enough :)

ShaharHD:
The 1st code snippet was problematic, as it seems it actually sends the data with the request (which can be problematic in some situations).

The 2nd code snippet is the one I might follow in the end.
Actually, this is how I actually had it in the 1st iteration of my code (to avoid the jshint warning) - but I assumed it has the same performance hit as if you define the function inside the for loop.

Moving forward, if you still believe there's no requirement to have the ability to store extra properties inside the request object itself, I'll close the pull request (and in the worst case scenario I will keep my branch alive and do catchup from time to time).

simov:
Yes, I don't think we need additional option here. As you can see you have plenty of options to implement that logic in your application layer. As for micro optimizations - don't worry about that too much. Performance issues can be identified only by appropriate tests.

ShaharHD:
Actually, only one option was viable - which was the last code snippet, and I'm still a bit reluctant - but will do for now :)

All the other ones were either sending unwanted data or creating instances of the method over and over again (which showed even in 8K requests) a major performance hit.

simov:
Closures are OK, calling function in JS have bare metal performance :)

ShaharHD:
Following your experience on this one :+1:

PR #1873, opened 2015-10-28
Exception fix for malformed urls to url.parse

cubesapp:
I am getting the following exception when calling request(URL,function(err,res,body){});

```
URIError: URI malformed at decodeURIComponent (native) at Url.parse (url.js:185:19) at Object.urlParse [as parse] (url.js:101:5) at Redirect.onResponse (/myproject/node_modules/request/lib/redirect.js:104:21) at Request.onRequestResponse (/myproject/node_modules/request/request.js:944:22) at ClientRequest.emit (events.js:95:17) at HTTPParser.parserOnIncomingClient [as onIncoming] (http.js:1692:21) at HTTPParser.parserOnHeadersComplete [as onHeadersComplete] (http.js:121:23) at Socket.socketOnData [as ondata] (http.js:1587:20) at TCP.onread (net.js:527:27)
```

The URL has multiple redirects.

Unfortunately, I don't know the exact URL, since we are processing 1000s of URLs at a time, and couldn't narrow it down.

Made the follow change to the redirect.js module for line 104:

``` js
try {
    request.uri = url.parse(redirectTo)
} catch(err) {
    return false;
}
```

Fixes #1864

Issue #1864, opened 2015-10-26
url.parse exception within redirect.js


cubesapp:
I am getting the following exception when calling request(URL,function(err,res,body){});

```
URIError: URI malformed at decodeURIComponent (native) at Url.parse (url.js:185:19) at Object.urlParse [as parse] (url.js:101:5) at Redirect.onResponse (/myproject/node_modules/request/lib/redirect.js:104:21) at Request.onRequestResponse (/myproject/node_modules/request/request.js:944:22) at ClientRequest.emit (events.js:95:17) at HTTPParser.parserOnIncomingClient [as onIncoming] (http.js:1692:21) at HTTPParser.parserOnHeadersComplete [as onHeadersComplete] (http.js:121:23) at Socket.socketOnData [as ondata] (http.js:1587:20) at TCP.onread (net.js:527:27)
```

The URL has multiple redirects.

Unfortunately, I don't know the exact URL, since we are processing 1000s of URLs at a time, and couldn't narrow it down.

Made the follow change to the redirect.js module for line 104:

``` js
try {
    request.uri = url.parse(redirectTo)
} catch(err) {
    return false;
}
```

Thoughts?

simov:
Obviously the parsing fails due to malformed URI, probably it's a good idea to have a try/catch block there. It's probably a redirect to absolute URL that's malformed, because we're resolving the relative ones on line 100.

cubesapp:
Created pull request #1873

PR #2114, opened 2016-03-04
Urlparse exception

rifflock:
This is the same as #1873 but with travis.ci fixes.

Fixes #1873

### DIFF
--- a/lib/redirect.js
+++ b/lib/redirect.js
@@ -104,7 +104,7 @@ Redirect.prototype.onResponse = function (response) {
 
   try {
     request.uri = url.parse(redirectTo)
-  } catch(e) {
+  } catch (e) {
     return false
   }
 
--- a/lib/redirect.js
+++ b/lib/redirect.js
@@ -101,11 +101,11 @@ Redirect.prototype.onResponse = function (response) {
   }
 
   var uriPrev = request.uri
-  
-  try{
-	  request.uri = url.parse(redirectTo)
+
+  try {
+    request.uri = url.parse(redirectTo)
   } catch(e) {
-	  return false;
+    return false
   }
 
   // handle the case where we change protocol from https to http or vice versa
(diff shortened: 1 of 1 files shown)


simov:
That's great, but we need a test for it :)

rifflock:
Okay, I'll see if I can get something reproducible put together. I'm pretty sure the issue occurs when the decode function gets half of a utf8 surrogate pair, so... I'll take a crack at it.

simov:
Yep, basically the description is here https://github.com/request/request/issues/1864 the test should be put in test-redirect.js I guess.

rifflock:
Welp, I'm at a bit of a loss.
I decided to use the "Pile of Poo" test from 
https://mathiasbynens.be/notes/javascript-unicode
And while I can reproduce the exception if I try to use `decodeURIComponent` on one of the surrogate pairs from the pile of poo directly in the test file. I can't get the same behavior by setting a `location` header that contains a half of a pile of poo (`\uD83D`), because it seems somewhere that invalid character is being replace with an `=` sign.
I'm going to keep working at this, but it looks like it will take me longer than I expected.

simov:
No problem, I'm kind of lost too, I though the error is about something else. Will take a look at it tomorrow.

simov:
That line seems to be found in [node v10](https://github.com/nodejs/node/blob/2f947e966588d46f85a49e387b8b107e575dff1d/lib/url.js#L185) and it's about basic auth string encoded in the url, but I'm not able to reproduce the bug using url.parse.

rifflock:
I'm not even entirely convinced that the `try-catch` around `url.parse` is the issue. Where I hit this issue in my code shouldn't be using any redirect logic. I just thought maybe I'd get lucky and this change would fix my issue.

I'm going to put `try-catch` blocks around my request calls in my code and see if it turns up anything. The error was pretty infrequent, maybe 15 times in 23000 calls. If I can get a more useful stack trace that points to something I'll open up a new pull request _with_ a functional test.

simov:
:+1: thanks @rifflock

Issue #2060, opened 2016-02-05
Redirects not redirecting


nmaxcom:
One simple POST request gets back a 302 redirection. I don't know if I should do something additional to get my script to follow that but... what I just get is that, a 302 and end of the script.

The documentation is not very extensive, and the examples are even more scarce. So I've been trying with no success adding `followRedirect: true, followAllRedirects: true` etc. When I use that last one param, the script can't even successfully log in to the site (with that POST).

The code goes something like this:

```
request({
            method: 'POST',
            followRedirect: true,
            uri   : Spider.globals.postURL,
            jar   : Spider.globals.jar,
            body  : Spider.globals.postMessage
        }, function(error, response, body) {
            console.log('Response status: ' + response.statusCode + ' -> ' + response.statusMessage);
            console.log(body);
        });
```

And the response:
********************
Trying to log in... success.
Response status: 302 -> Found
<html><body>You are being <a href="https://xxx">redirected</a>.</body></html>

DoctorBud:
I am having problems that might be related to this issue.

I'm trying to POST data to a server that responds with a 301. I use the followAllRedirects:true option, and `request` detects the 301 and initiates a redirect. However, for some reason (a bug, IMHO) it converts the POST request to a GET, which results in the redirect target receiving no POST data and therefore failing to deliver a useful response.

I think there is a bug in lib/redirect.js here: https://github.com/request/request/blob/master/lib/redirect.js#L116-L119

```
  if (self.followAllRedirects && request.method !== 'HEAD'
    && response.statusCode !== 401 && response.statusCode !== 307) {
    request.method = 'GET'
  }
```

I think that the `request` library needs to have the option of redirecting POST and preserving the original method and body, rather than converting to GET. If you read the `curl` manual, it has an option `--post301` with the following documentation that might be relevant here:

>  (HTTP)  Tells  curl to respect RFC 2616/10.3.2 and not convert POST requests into GET
              requests when following a 301 redirection. The non-RFC behaviour is ubiquitous in web
              browsers,  so curl does the conversion by default to maintain consistency. However, a
              server may require a POST to remain a POST after such a redirection. This  option  is
              meaningful only when using -L, --location (Added in 7.17.1)

And here's the relevant text from RFC 2616/10.3.2 at https://tools.ietf.org/html/rfc2616#section-10.3.2:

> Note: When automatically redirecting a POST request after receiving a 301 status code, some existing HTTP/1.0 user agents will erroneously change it into a GET request.

[...]

PR #2137, opened 2016-03-24
Honour RFC 1945 & RFC 2048

ThanosSiopoudis:
RFC 1945 & RFC 2048 state that upon a 301 or 302 redirect, user agents should not automatically change the request method to GET:

> Note: When automatically redirecting a POST request after
>        receiving a 301 status code, some existing user agents will
>        erroneously change it into a GET request.

(There is a relevant note for the 302 redirection as well)
https://tools.ietf.org/html/rfc1945#section-9.3

This relates to #2060

### DIFF
--- a/lib/redirect.js
+++ b/lib/redirect.js
@@ -114,14 +114,16 @@ Redirect.prototype.onResponse = function (response) {
     }
   )
   if (self.followAllRedirects && request.method !== 'HEAD'
-    && response.statusCode !== 401 && response.statusCode !== 307) {
+    && response.statusCode !== 401 && response.statusCode !== 307
+    && response.statusCode !== 302 && response.statusCode !== 301) {
     request.method = 'GET'
   }
   // request.method = 'GET' // Force all redirects to use GET || commented out fixes #215
   delete request.src
   delete request.req
   delete request._started
-  if (response.statusCode !== 401 && response.statusCode !== 307) {
+  if (response.statusCode !== 401 && response.statusCode !== 307
+    && response.statusCode !== 301 && response.statusCode !== 302) {
     // Remove parameters from the previous response, unless this is the second request
     // for a server that requires digest authentication.
     delete request.body


FredKSchott:
Tests are failing, and this has gotten stale enough that merge conflicts need to be addressed.

>   If the 301 status code is received in response to a request using
>   the POST method, the user agent must not automatically redirect the
>   request unless it can be confirmed by the user, since this might
>   change the conditions under which the request was issued.

I agree that the current behavior is wrong, but redirecting as a POST is potentially worse. I would love to accept a PR to not redirect POSTs at all if you have the time.

FredKSchott:
Thank you for including the spec btw, I wouldn't have been able to review this without it :)

mikeal:
wait, are we actually changing the method to GET? I thought we *never* did that under the assumption that we needed to offer something lower level so that higher order libraries could customize how to interpret the spec?

mikeal:
my memory may be failing me, or perhaps I just never noticed this land, but we should consider backing that out.

it's part of the spec but because of the layer we operate at we can't make assumptions about the intended behavior or our users when they are attempting to send a POST or PUT body. If that turns into a GET then their data isn't posted and we will be reporting a successful request with a 200 response which they will likely misinterpret.

ThanosSiopoudis:
Exactly. Redirecting with a GET regardless of the requested method, does lead to misinterpretations. In reality, per the spec, it should either not follow the redirect and simply return it to the user to let him decide what to do with it, or add an option to either follow redirects with the original method, or not follow it at all.

If we can decide on how to approach this, I'm happy to clean this up and add any needed tests.

FredKSchott:
@ThanosSiopoudis Can you rebase/merge the PR with master and see if this is still needed? A `followOriginalHttpMethod` option was added (possibly to address this) that isn't showing up in the PR diff.

FredKSchott:

PR #2164, opened 2016-04-11
Catch errors from the underlying http module

### LINKED ISSUE #2120: request.get crashes with "The header content contains invalid characters"
Try this:

```
require('request').get('http://www.test.com/אבגד.pdf');
```

After a few milliseconds this crashes with `TypeError('The header content contains invalid characters')`

Now if I wrap it in `encodeUR` - it succeeds.
But if we receive urls from an external source - they may or may not be encoded. Is there some kind of built in mechanism to deal with this? (Other than doing `encodeURI(decodeURI(url))`)

simov:
Fixes #2120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,19 @@
 language: node_js
+
 node_js:
   - node
-  - io.js
   - 0.12
   - 0.10
-sudo: false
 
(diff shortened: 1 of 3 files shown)


danielgindi:
:+1:

Meabed:
👍

khurrumqureshi:
👍

simov:
Version 2.71 is published with the fix.

PR #2165, opened 2016-04-12
Check for self.req existence inside the write method

### LINKED ISSUE #2120: request.get crashes with "The header content contains invalid characters"

simov:
Fixes #2120

Also see https://github.com/request/request/issues/2120#issuecomment-208954497 and https://github.com/request/request/issues/2164

### DIFF
--- a/request.js
+++ b/request.js
@@ -1382,7 +1382,9 @@ Request.prototype.write = function () {
   if (!self._started) {
     self.start()
   }
-  return self.req.write.apply(self.req, arguments)
+  if (self.req) {
+    return self.req.write.apply(self.req, arguments)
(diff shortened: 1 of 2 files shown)


khurrumqureshi:
:+1:

PR #2166, opened 2016-04-12
Add 'delete' alias for the 'del' API method

### LINKED ISSUE #1917: Just a little api critic.
Hi guys,

i really like this library, you guys made an awesome job. I like to use this library.
But, please provide a alias for request.del => request.delete, because i was confused, because the method name mostly represents the http method and that surly does sense to me. 

Best,
mrlove

MuhanZou:
Only Using 'del' seems to be a little bit opinionated if there is no 'delete' method provided together; and 'del' also introduced redundant ternary assignment. Thanks!

Fixes #1917

### DIFF
--- a/README.md
+++ b/README.md
@@ -888,6 +888,14 @@ Same as `request()`, but defaults to `method: "HEAD"`.
 request.head(url)
 ```
 
+### request.delete
+
+Same as `request()`, but defaults to `method: "DELETE"`.
+
+```js
+request.delete(url)
+```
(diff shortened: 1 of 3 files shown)


simov:
@MuhanZou can you pull https://github.com/simov/request/commit/865e981f2e36d79ef07a3c44f8e3470e6a7b89f2 here? It's safer to define the methods separately. I also added quotes around the property for now.

PR #2167, opened 2016-04-13
Fix TravisCI badge reference master branch

a0viedo:
I also noticed there're two badges for Coverage, one pointing to Codecov and the other one to Coveralls. Is this intended or one should not be used?


simov:
I guess that was the reason why the icons was stuck on *build error*?

Yes we need both of those coverage badges. Coveralls was not showing up the file coverage at some point, something was bugged out with the request repo. We tried just about everything and the issue is still open and not resolved on the Coveralls issue tracker.

Issue #2171, opened 2016-04-15
Request does not support 'deflate' encoding.


czardoz:
This request results in the raw response being dumped to the console:
```javascript
var req = require('request');

req({
    url: 'http://httpbin.org/deflate',
    method: 'GET',
    headers: {},
    followAllRedirects: false,
    followRedirect: true,
    jar: true,
    timeout: 15000,
    gzip: true,
    rejectUnauthorized: false,
    strictSSL: true
}, function (err, res, body) {
    if (err) {
        console.error(err.stack || err);
    }
    console.log(body);
});
```

The expected behavior is that `body` contains the decoded response.

czardoz:
Changing the gzip handling to 
```javascript
      if (contentEncoding === 'gzip') {
        responseContent = zlib.createGunzip()
        response.pipe(responseContent)
      } else if (contentEncoding === 'deflate') {
        responseContent = zlib.createInflate()
        response.pipe(responseContent)
      }
```
in [request.js](https://github.com/request/request/blob/5a8ddc8102a0897ca1aacf222576258ad87105c1/request.js#L928) ensures that deflate responses work.

I'm willing to submit a PR for this change, but are there any other considerations here?

simov:
That's weird, I though we support it already. I'm wondering if the `gzip` option would be appropriate in this case though.

czardoz:
True, `gzip` is probably not the right option. But I don't think it is nice to add a new option for every new encoding that comes up either.

One solution is to use the `gzip` option, since `deflate` and `gzip` [are related](https://tools.ietf.org/html/rfc1951). Not sure if this is backwards compatible though.

Another option is to include `deflate` as a new option, this will lead to a bit of code duplication, and too many options, I think. 

I'm okay with either of the above options though (or any that I might not have thought of) :)

simov:
Yep let's keep it under `gzip` for now. I have a better approach for handling the options [here](https://github.com/request/core/blob/master/lib/options/gzip.js).

Don't forget to add a test in the `test-gzip` file.

czardoz:
Done!

Issue #2045, opened 2016-01-29
HEAD requests with Node v5 have issues with gzipped content and fail.


czardoz:
This script reproduces the issue:

````js    
    var req = require('request');
    
    req({
        url: 'http://google.com',
        method: 'HEAD',
        headers: {},
        followAllRedirects: false,
        followRedirect: true,
        jar: true,
        timeout: 15000,
        gzip: true,
        rejectUnauthorized: false,
        strictSSL: true
    }, function (err, res, body) {
        if (err) {
            console.error(err.stack || err);
        }
    });
```

Steps to reproduce:

    aniket [/tmp/request-bug] -> node -v
    v5.1.0
    aniket [/tmp/request-bug] -> npm -v
    3.3.12
    aniket [/tmp/request-bug] -> cat node_modules/request/package.json| grep version
      "version": "2.69.0"
    aniket [/tmp/request-bug] -> node reproduce.js 
    Error: unexpected end of file
        at Zlib._handle.onerror (zlib.js:363:17)
    aniket [/tmp/request-bug] ->

nathanmarks:
@czardoz FWIW I had the same issue with some HEAD requests and it is because of `followAllRedirects`. Unless you set it to `true`, it won't follow HEAD redirects (IE non www to www, or http to https).

czardoz:
Is this issue reproducible for others? 

@simov Is there a label for a "confirmed" bug? I can consistently reproduce it with Node v5.

nathanmarks:
@czardoz Not sure if this helps you, but I was doing `HEAD` requests to avoid downloading entire responses.

I ended up switching to `GET` requests, streaming the response and aborting the request early.

czardoz:
@nathanmarks While that's a viable workaround, I use this library to run user-specified HTTP requests, so I cannot really change the request method :)

simov:
Fixed here #2176

Let me know what do you think.

PR #2172, opened 2016-04-15
Add support for deflate content encoding

czardoz:
Fixes https://github.com/request/request/issues/2171


simov:
Looking good, will merge it shortly :+1:

czardoz:
Is this also a good time to fix https://github.com/request/request/issues/2045?

PR #2176, opened 2016-04-16
Do not try to pipe Gzip responses with no body

### LINKED ISSUE #2045: HEAD requests with Node v5 have issues with gzipped content and fail.

### LINKED ISSUE #2177: Handling of empty responses when dealing with GZip
This is a continuation of the discussion in https://github.com/request/request/pull/2176

Basically, what should `request` do when a server returns an empty response with the `Content-Encoding` header set to `gzip`?

simov:
Fixes #2045
Fixes #2177

### DIFF
--- a/request.js
+++ b/request.js
@@ -920,8 +920,20 @@ Request.prototype.onRequestResponse = function (response) {
       self._ended = true
     })
 
+    var noBody = function (code) {
+      return (
+        self.method === 'HEAD'
+        // Informational
+        || (code >= 100 && code < 200)
+        // No Content
+        || code === 204
+        // Not Modified
+        || code === 304
+      )
+    }
+
     var responseContent
-    if (self.gzip && self.method !== 'HEAD') {
+    if (self.gzip && !noBody(response.statusCode)) {
       var contentEncoding = response.headers['content-encoding'] || 'identity'
       contentEncoding = contentEncoding.trim().toLowerCase()
 
--- a/request.js
+++ b/request.js
@@ -921,7 +921,7 @@ Request.prototype.onRequestResponse = function (response) {
     })
 
     var responseContent
-    if (self.gzip) {
+    if (self.gzip && self.method !== 'HEAD') {
       var contentEncoding = response.headers['content-encoding'] || 'identity'
       contentEncoding = contentEncoding.trim().toLowerCase()
 
--- a/tests/test-gzip.js
+++ b/tests/test-gzip.js
@@ -21,6 +21,14 @@ var server = http.createServer(function(req, res) {
     res.end()
     return
   }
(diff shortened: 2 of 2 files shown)


request.js
```diff
@@ -921,7 +921,7 @@ Request.prototype.onRequestResponse = function (response) {
     })
 
     var responseContent
-    if (self.gzip) {
+    if (self.gzip && self.method !== 'HEAD') {
```
czardoz: This will fix the particular issue with HEAD requests, but will still cause issues if the response is empty for say a GET. A better way of handling would be to catch the Error from zlib.

czardoz:
Thanks, @simov!

request.js
simov: Figuring out if the response is empty beforehand would be great, will think about it. I would rather not use `try/catch` if there is some other way to detect the condition.

simov: Unfortunately there is no reliable way to determine upfront if the response body is empty. And in fact the error is thrown because the gzip content is invalid which may happen even if you have a body.
So at least we can do is forward the error events so that the user can handle them accordingly. Which means that you will receive an error when the request method is GET, the gzip option is true, and there is no response body.

simov: You can make assumption in your own layer, that when you receive that error under certain circumstances it may mean something, but as far as request is concerned the error should be forwarded.
I'm open for suggestion though.

simov: Wait, it's already like that, I got mislead.

simov: Anyway still this PR fixes the mentioned issue in the description which is the most common use case.

@czardoz you can open up another issue and point to this one, so we can have a discussion there.

czardoz: Done: #2177

czardoz:
LGTM! :cake:

simov:
Version 2.72 is out! :sparkles:

PR #2205, opened 2016-05-07
Use server-destory to close hanging sockets

simov:
when running the coverage report (in single process)


simov:
No more 10 minute builds.

New record - 3 minutes :sparkles:

PR #2221, opened 2016-05-24
Adding a simple Response object reference in argument specification

calamarico:
When i read request arguments specification for first time, i didn't recognize that IncomingMessage is created by ClientRequest and passed as the first argument 'request' or 'response', i think this reference in readme improve this great package :D


FredKSchott:
Thanks @calamarico!

PR #2249, opened 2016-06-17
Fix target.method of defaults.del

katemihalikova:
`target.method` is set to `DEL` instead of `DELETE` in `defaults.del`.

Introduced in #2175 (2ef974a), broken since 2.72.0.

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,10 +115,11 @@ request.defaults = function (options, requester) {
 
   var defaults      = wrapRequestMethod(self, options, requester)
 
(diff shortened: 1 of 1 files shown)


simov:
Hi @katemihalikova can you add a failing test for this?

katemihalikova:
After more investigation it turned out that even though [here](https://github.com/request/request/blob/master/index.js#L95) `target.method` is set to `DEL` it is later replaced with `DELETE` [here](https://github.com/request/request/blob/master/index.js#L62). The original problem was found in our own codebase, not in request. Thanks anyway.

PR #2282, opened 2016-07-13
Missing semicolons in example

grubersjoe:
Two semicolons were missing in the example


simov:
Hi, @grubersjoe, thanks for stopping by but unfortunately we don't use semicolons any more.

Issue #2315, opened 2016-08-03
self.start invoked when request failed.


calidion:
```
TypeError: self.start is not a function
  at Socket.Request.write [as _write] (/node_modules/request/request.js:1380:10)
  at doWrite (_stream_writable.js:300:12)
  at clearBuffer (_stream_writable.js:407:7)
  at Socket.Writable.uncork (_stream_writable.js:237:7)
  at ClientRequest._flushOutput (_http_outgoing.js:673:10)
  at ClientRequest.OutgoingMessage._flush (_http_outgoing.js:647:16)
  at _http_client.js:156:10
  at callSocketMethod (_http_client.js:531:7)
  at ClientRequest.onSocket (_http_client.js:536:7)
  at ClientRequest.g (events.js:260:16)
  at emitOne (events.js:82:20)
  at ClientRequest.emit (events.js:169:7)
  at tickOnSocket (_http_client.js:502:7)
  at onSocketNT (_http_client.js:514:5)
  at nextTickCallbackWith2Args (node.js:442:9)
  at process._tickDomainCallback (node.js:397:17)
```

w446108264:
Did you solve it？

w446108264:
I use "request": "^2.74.0" on the my main project, but a lib which use "request": "^2.69.0". it's not ok. catch like that TypeError: self.start is not a function.

so i fork the lib and modify as "request": "^2.74.0" , it works 😂

calidion:
when will this be solved?!

calidion:
@w446108264 
I don't have that luck.
all my modules dependent on this module are ruined.
none of then is able to work properly.

w446108264:
😭

开始搜索到这里，以为是个外国人，刚好把问题解决了，就回复一下，没想到你讲中文。

更没想到的是，导致我项目问题的那个lib仓库是你的。
https://github.com/node-weixin/node-weixin-message

https://github.com/node-weixin/node-weixin-message/blob/master/package.json
可见"request": "^2.69.0"   我把他修改为 2.74.0，并暂时删除"node-weixin-request": "^0.4.0"等

项目的问题暂时解决了。本来打算pull request的，因为解决的粗糙还没动手。
见(https://github.com/w446108264/node-weixin-message),

另外node-weixin-message部分方法已经过时了，微信企业号现在似乎（刚接触微信玩法）把内容都加密为Encrypt。
https://github.com/node-weixin/node-weixin-message/blob/master/lib/incoming.js
原来方法

```js

[...]

PR #2316, opened 2016-08-03
Not to start when object is not a Request instance

calidion:
to fixed bug https://github.com/request/request/issues/2315


FredKSchott:
If you do this, other bugs will occur. The real problem is that `write()` is being called on the wrong context. I would accept a PR to throw a helpful error here, but silently failing will just cover the problem up more.

This is an old PR so I'll close it now in case you have moved on, but if you (or anyone else) are interested in creating a new PR to fix this with a helpful error message that would be much appreciated.

PR #2346, opened 2016-08-25
Progress event

### LINKED ISSUE #1450: Implement progress events
It would be nice to have (optional) progress events that would fire for progress (against content-length if available) on the raw entity body.

There are 2 libraries that do progress events for request, [request-progress](https://www.npmjs.com/package/request-progress) and [progress-stream](https://www.npmjs.com/package/progress-stream).

However both of them fail when you use `gzip: true`. `Content-Length` describes the length of the entity body (compressed) but both of these libraries only have access to the inflated (uncompressed) stream request gives out.

It doesn't seem possible for these libraries to dig themselves into request's internals to access the raw stream and watch the data that streams through it. So support for progress events looks like something that can only be implemented correctly from within request.

scotttrinh:
This completes my programs requirements, but as I'm not familiar with all of the ins and outs of `request`, I'm sure I'm missing some other point where a progress event might make sense.

At the moment it:
- ⏲  Adds a transferred property to the `Request` object to track bytes sent. 
- 🛫  During the `write` operation, it will emit a `progress` event that includes the standard `ProgressEvent` object as defined in the [ProgressEvent spec](https://xhr.spec.whatwg.org/#interface-progressevent).

Again, due to my unfamiliarity with the code base, and being the first time I've used `tape` tests, I would say my test cases are pretty minimal and permissive, so I definitely welcome some direction from the team on that.

Closes #1450

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,11 @@ request.post({url:'http://service.com/upload', formData: formData}, function opt
     return console.error('upload failed:', err);
   }
   console.log('Upload successful!  Server responded with:', body);
+}).on('progress', function(progress) {
+  if (progress.lengthComputable) {
+    // Follows the format of ProgressEvent
+    // See: https://xhr.spec.whatwg.org/#interface-progressevent
+  }
 });
 ```
 
--- a/request.js
+++ b/request.js
@@ -120,6 +120,7 @@ function Request (options) {
   if (options.method) {
     self.explicitMethod = true
   }
+  self.transferred = 0;
   self._qs = new Querystring(self)
   self._auth = new Auth(self)
   self._oauth = new OAuth(self)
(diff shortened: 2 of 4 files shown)


scotttrinh:
I'll dig into the failures for the older Node platforms, but does anyone have any off-the-cuff thoughts about what might be going on?

scotttrinh:
image

`form-data` 1.0.0 was published 6 hours ago, so this broke in CI and not locally until I updated. They changed the order of lookup which changed the error string we were searching for in our unit tests. Since this seems a bit brittle, I've updated the test to check that it throws a `TypeError` instead of being tied to the error string.

devinterx:
In what status there is an acceptance of this pull request? @simov

FredKSchott:

3. [ ] Push your updated branch to Github to update this PR.
4. [ ] Make sure this PR's Travis tests are still passing. (Run `npm test` locally if tests are still failing).
5. [ ] Comment here to let us know that you've updated the PR and are still interested in getting it merged. Somebody from the project will respond and work with you to review.

doughock:
I would be interested in a progress indicator for formData uploads with attached files.

PR #2352, opened 2016-08-28
Add responseStartTime timing

paulirish:
When `time:true` is set in options, elapsedTime is added to the response object.

This PR will _also_ include `responseStartTime`, the timestamp when the response begins. With this moment tagged, we can now recreate a rich network waterfall of the network activity. 

For example, here's a network waterfall of some project's `npm install` with this patch and [`request-capture-har`](https://github.com/paulirish/node-request-capture-har):
image

### DIFF
--- a/request.js
+++ b/request.js
@@ -958,6 +958,10 @@ Request.prototype.onRequestResponse = function (response) {
     })
 
     responseContent.on('data', function (chunk) {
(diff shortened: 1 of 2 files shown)


paulirish:
Confirmed that the build is green once the fix in #2349 is applied.

simov:
:+1:

PR #2378, opened 2016-09-17
Update form-data dependency

mscdex:
This should decrease the total footprint of `request` by ~5.5MB.


mscdex:
Ah, looks like the greenkeeper bot beat me to it in https://github.com/request/request/pull/2377. However, it might be nice to use `^` instead of `~` for the version since `form-data` seems to follow semver?

alexindigo:
Looks like `test-pipes` isn't that stable. Greenkeeper didn't pass.

simov:
:+1:

PR #2410, opened 2016-10-11
Clear timeout on `#abort()`

elmigranto:
Calling `#abort()` does not clear timeout, process hangs until timer is triggered. There are couple of places this is done correctly, `#abort()` happens not to be one of them. There are couple of relevant issues and PRs from **years ago**.

Sample script to reproduce:

``` js
'use strict';
const request = require('request');
const started = Date.now();
const printTime = label => () => console.log('%d ms\t%s', Date.now() - started, label);

const req = request({
  uri: 'http://example.com',
  timeout: 2500
}, printTime('callback'));

setTimeout(() => {
  req.abort();
  printTime('aborted')();
}, 10);

process.on('exit', printTime('exit'));

// 27 ms    aborted
// 2519 ms  callback
// 2521 ms  exit
```

Let's close this one finally. Can any of you guys help, please? @mikeal @simov

### DIFF
--- a/request.js
+++ b/request.js
@@ -806,10 +806,7 @@ Request.prototype.onRequestError = function (error) {
     self.req.end()
     return
   }
-  if (self.timeout && self.timeoutTimer) {
-    clearTimeout(self.timeoutTimer)
-    self.timeoutTimer = null
-  }
+  self.clearTimeout()
   self.emit('error', error)
 }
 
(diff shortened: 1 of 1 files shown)


FredKSchott:
@elmigranto still interested in getting this merged? happy to help review if so.

elmigranto:
@fredkschott ofc, thanks.

request.js
```diff
@@ -1426,6 +1421,13 @@ Request.prototype.destroy = function () {
   }
 }
 
+Request.prototype.clearTimeout = function () {
+  if (this.timeout && this.timeoutTimer) {
```
FredKSchott:
```diff
-  if (this.timeout && this.timeoutTimer) {
+  if (this.timeoutTimer) {
```

We should call `clearTimeout(this.timeoutTimer)` anytime `this.timeoutTimer` was created, even if `this.timeout` wasn't passed.

FredKSchott:
+1 great catch, thanks for fixing and refactoring a bit.

Can you add a test as well so that we don't regress? Just checking for `timeoutTimer` after calling `abort()` is fine.

request.js
reconbot: agreed

PR #2414, opened 2016-10-14
Improve test-timeout reliability

mscdex:
This commit should allow the 'connect timeout' test to pass more often than not.


mscdex:
@simov Does this LGTY? I'm still encountering this locally occasionally. I have also seen it on Travis before FWIW.

PR #2424, opened 2016-10-16
Handle buffers directly instead of using "bl"

zertosh:
`bl` does a lot, but `request`'s needs are pretty simple. This PR removes `bl` in favor of operating on buffers directly. There is no change in functionality here - `bl` also calls `Buffer.concat` when slicing an all buffer `BufferList`.

### DIFF
--- a/request.js
+++ b/request.js
@@ -6,7 +6,6 @@ var http = require('http')
   , util = require('util')
   , stream = require('stream')
   , zlib = require('zlib')
-  , bl = require('bl')
   , hawk = require('hawk')
   , aws2 = require('aws-sign2')
   , aws4 = require('aws4')
@@ -1005,14 +1004,16 @@ Request.prototype.onRequestResponse = function (response) {
 Request.prototype.readResponseBody = function (response) {
   var self = this
   debug('reading response\'s body')
-  var buffer = bl()
+  var buffers = []
+    , bufferLength = 0
     , strings = []
 
   self.on('data', function (chunk) {
-    if (Buffer.isBuffer(chunk)) {
-      buffer.append(chunk)
-    } else {
+    if (!Buffer.isBuffer(chunk)) {
       strings.push(chunk)
+    } else if (chunk.length) {
+      bufferLength += chunk.length
+      buffers.push(chunk)
     }
   })
   self.on('end', function () {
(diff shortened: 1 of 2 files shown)


mikeal:
If i recall, the original thinking was "maybe we can eventually just pass the bl instance to the callback" instead of doing the concatenation and underlying memcopy of the buffers.

Turns out it's just too difficult to "pretend" to be a Buffer reliably and integrate with instanceof checks and `Buffer.isBuffer()`.

mikeal:
To be clear, I'm +1 on the premise, but I haven't had the time to code review quite yet.

zertosh:
Thank you, I appreciate the background.

For context, this is just one of many more upcoming PRs. My end goal is to improve `request`'s load time. Right now it's taking ~90-100ms - which is not a trivial amount. There are common cases in [popular](https://github.com/yarnpkg/yarn/pull/441) [projects](https://github.com/tmpvar/jsdom/pull/1630) where `request` is in the load path, but it's not used every time. I'm investigating other optimizations like lazy loading request's deps and lazily instantiating the various support classes.

mikeal:
I wonder why the code coverage went down, this still hit a bunch of our existing tests :(

mscdex:
This could be improved by keeping a running total that's passed as the second argument to `Buffer.concat()`. It saves time by not having `Buffer.concat()` loop through all of the Buffers to find the total length.

zertosh:
I've updated the PR by rebasing onto master and adding the length tracking as suggested by @mscdex

request.js
```diff
@@ -1021,22 +1022,20 @@ Request.prototype.readResponseBody = function (response) {
       debug('aborted', self.uri.href)
       // `buffer` is defined in the parent scope and used in a closure it exists for the life of the request.
       // This can lead to leaky behavior if the user retains a reference to the request object.
-      buffer.destroy()
+      buffers = []
```
mscdex: Minor nit: perhaps to be consistent with the changes down below, `bufferLength = 0` should be added after this line as well?

zertosh: done!

mscdex:
LGTM and CI is green :-)

simov:
@zertosh :+1: 
@mscdex you have a pending invitation to the request team from a long time ago, let me know if you want to send you another one :+1:

PR #2415, opened 2016-10-14
Re-enable timeout tests on Travis + other fixes

mscdex:
What started out as a simple re-enabling of `tests/test-timeout.js` on Travis has turned into more cleanup and fixes along the way.

If you want any of these commits parted out into separate PRs, let me know.

I have tested these commits on Travis many times now and each time all 4 node targets have been green, so I am fairly confident not only about `tests/test-timeout.js`, but also flaky tests in `tests/test-pipes.js`, and other random failures.

One possibly "major" change included here is the change in timeout behavior. IMHO the timeout changes (in request.js) I've provided here more accurately reflect the intended behavior.

PR #2431, opened 2016-10-19
Improve timeouts accuracy and node v6.8.0+ compatibility

mscdex:
This PR fixes a timeout-related compatibility issue with node v6.8.0+ and makes the connection and request/response timeout more accurate by starting them at more appropriate times.


mscdex:
FWIW the CI failure here is fixed by the relevant commit in https://github.com/request/request/pull/2415.

PR #2435, opened 2016-10-25
Add followOriginalHttpMethod to redirect to original HTTP method

### LINKED ISSUE #2118: Post JSON & followAllRedirects ends with GET method
I'm trying to consume a staging web service using https protocol.
The POST receive a 301 redirect, which I tried to make request following silently using (followAllRedirects).
curl hasn't any problem with that
      
      * Connected to api.staging.eu.xxx.com (x.x.x.x) port 443 (#0)
      * TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
      * Server certificate: *.staging.eu.xxx.com
      * Server certificate: RapidSSL SHA256 CA - G3
      * Server certificate: GeoTrust Global CA
      > POST /api/v1/employees/ HTTP/1.1
      > Host: api.staging.eu.xxx.com
      > User-Agent: curl/7.43.0
      > Content-type: application/json
      > Accept: application/json
      > X-API-KEY: xxxxxxxxx
      > Content-Length: 502

request fails there, because of redirect.js

To debug it and get it working to be sure, I just removed the following lines:

      if (self.followAllRedirects && request.method !== 'HEAD'
        && response.statusCode !== 401 && response.statusCode !== 307) {
        // request.method = 'GET'                                     // <= removed to debug
      }
      // request.method = 'GET' // Force all redirects to use GET || commented out fixes #215
      delete request.src
      delete request.req
      delete request._started
      if (response.statusCode !== 401 && response.statusCode !== 307) {
        // Remove parameters from the previous response, unless this is the second request
        // for a server that requires digest authentication.
        // delete request.body                                        // <= removed to debug
        delete request._form
        if (request.headers) {
          // request.removeHeader('host')                             // <= removed to debug
          // request.removeHeader('content-type')                     // <= removed to debug
          // request.removeHeader('content-length')                   // <= removed to debug

I saw a few issues which seems refer to the redirect (#556, #215...), but I'm absolutely not an expert of SSL &  http redirection, so, not sure about what should be done or not

kirrg001:
Hey 👋

closes #2118

You can set `followOriginalHttpMethod` to follow the original HTTP method.
The problem was that request followed the requests with a hardcoded HTTP method.

---

An alternative solution would be to always redirect to the original HTTP method. Tell me what do you prefer and i will update my PR 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -762,6 +762,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 
 - `followRedirect` - follow HTTP 3xx responses as redirects (default: `true`). This property can also be implemented as function which gets `response` object as a single argument and should return `true` if redirects should continue or `false` otherwise.
 - `followAllRedirects` - follow non-GET HTTP 3xx responses as redirects (default: `false`)
+- `followOriginalHttpMethod` - by default we redirect to HTTP method GET. you can enable this property to redirect to the original HTTP method (default: `false`)
 - `maxRedirects` - the maximum number of redirects to follow (default: `10`)
 - `removeRefererHeader` - removes the referer header when a redirect happens (default: `false`). **Note:** if true, referer header set in the initial request is preserved during redirect chain.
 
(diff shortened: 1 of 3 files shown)


simov:
Thanks @kirrg001 very clean work :+1:

gabmontes:
Thanks guys!!!

PR #2447, opened 2016-11-03
Always set request timeout on keep-alive connections

mscdex:


mscdex:
The changes in 001eae3d479a13350b06b8d3f8182b2060a2c60d would cause the http request timeout to only get set if a new socket was being used. This PR makes sure it gets set even when we're already connected.

simov:
I can publish another patch no problem, I had to ask you first. So is this good for merging/publishing?

mscdex:
It should be good to go now, I've added a test for it.

simov:
Thank you! Published.

PR #2452, opened 2016-11-06
Adds .timings array with DNC, TCP, request and response times

nicjansma:
This improves upon https://github.com/request/request/pull/1459 by adding a timings breakdown for the following phases:

* DNS
* TCP
* Request
* Download

First, instead of relying on `new Date().getTime()` we can use `process.hrtime()` to get a higher-resolution monotonically non-decreasing timestamp.  This is utilized via the [`performance-now`](https://github.com/myrne/performance-now) package.

The previous PR tracked `startTime` (not exposed) and `elapsedTime` (exposed on the request).  This PR adds a new `.timings` object that adds new timestamps and elapsed times.

Note the "timestamps" in this object are not Unix-epoch based like `new Date.getTime()` -- they are `milliseconds.microsecond` resolution relative to when the `performance-now` module got loaded.

If there is a redirect when fetching the resource, the timestamps are only for the final resource fetched.

Timestamps:

* `start`: Timestamp when `request()` was initialized
* `socket` Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires.  This happens when the socket is assigned to the request (after DNS has been resolved).
* `connect`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_connect) module's `connect` event fires.  This happens when the server acknowledges the TCP connection.
* `response`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_response) module's `response` event fires.  This happens when the first bytes are received from the server.
* `end`: Timestamp when the last bytes of the response are received.

Durations:

* `dns`: Duration of DNS lookup (`timings.socket` - `timings.start`)
* `tcp`: Duration of TCP connection (`timings.connect` - `timings.socket`)
* `firstByte`: Duration of HTTP server first byte (`timings.response` - `timings.connect`)
* `download`: Duration HTTP fetch (`timings.end` - `timings.response`)
* `total`: Duration entire HTTP round-trip (`timings.end` - `timings.start`)

Example:

```
var req = request({ url: "http://foo.com", time: true }, function(err, response, body) {
  console.log("elapsedTime", response.elapsedTime);
  console.log(response.timings);
});

/*
elapsedTime 271
{ start: 8.496343,
  socket: 11.571675,
  connect: 61.353872,
  response: 102.523232,
  end: 279.228004,
  dns: 3.0753320000000013,
  tcp: 49.782197000000004,
  firstByte: 41.16935999999999,
  download: 176.704772,
  total: 270.731661 }
*/
```


mikeal:
First of all, this is great.

The only problem I see is the additional dependency. We've actually been combing through our deps and removing deps that are this small in order to improve the load time of request which is currently bogged down in loading a lot of relatively small deps that could be inlined.

README.md
```diff
@@ -812,7 +812,17 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
...
+  - `start`: Timestamp when `request()` was initialized
+  - `socket` Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires.  This happens when the socket is assigned to the request (after DNS has been resolved).
+  -  `connect`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_connect) module's `connect` event fires.  This happens when the server acknowledges the TCP connection.
+  - `response`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_response) module's `response` event fires.  This happens when the first bytes are received from the server.
+  - `end`: Timestamp when the last bytes of the response are received.
+  - `dns`: Duration of DNS lookup (`timings.socket` - `timings.start`)
+  - `tcp`: Duration of TCP connection (`timings.connect` - `timings.socket`)
+  - `download`: Duration HTTP fetch (`timings.end` - `timings.response`)
+  - `total`: Duration entire HTTP round-trip (`timings.end` - `timings.start`)
```
Huachao: also add `firstByte` description

tests/test-timing.js
```diff
@@ -31,8 +31,19 @@ tape('non-redirected request is timed', function(t) {
...
+    t.equal(typeof res.timings, 'object')
     t.equal((res.elapsedTime > 0), true)
     t.equal((res.responseStartTime > r.startTime), true)
+    t.equal((res.timings.start > 0), true)
+    t.equal((res.timings.socket >= res.timings.start), true)
+    t.equal((res.timings.connect >= res.timings.socket), true)
+    t.equal((res.timings.response >= res.timings.connect), true)
+    t.equal((res.timings.end >= res.timings.response), true)
+    t.equal((res.timings.dns >= 0), true)
+    t.equal((res.timings.tcp > 0), true)
```
Huachao: `>` or `>=` ?

README.md
```diff
@@ -812,7 +812,17 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
...
+  - `start`: Timestamp when `request()` was initialized
+  - `socket` Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires.  This happens when the socket is assigned to the request (after DNS has been resolved).
+  -  `connect`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_connect) module's `connect` event fires.  This happens when the server acknowledges the TCP connection.
+  - `response`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_response) module's `response` event fires.  This happens when the first bytes are received from the server.
+  - `end`: Timestamp when the last bytes of the response are received.
+  - `dns`: Duration of DNS lookup (`timings.socket` - `timings.start`)
+  - `tcp`: Duration of TCP connection (`timings.connect` - `timings.socket`)
+  - `download`: Duration HTTP fetch (`timings.end` - `timings.response`)
```
Huachao: change to `Duration of HTTP`?

README.md
```diff
@@ -812,7 +812,17 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
 
 ---
 
-- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution, and the result provided on the response's `elapsedTime` property. The `responseStartTime` property is also available to indicate the timestamp when the response begins.
+- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution, and the result provided on the response's `elapsedTime` property. The `responseStartTime` property is also available to indicate the timestamp when the response begins.  In addition, there is a `.timings` object available with the following properties:
```
Huachao: remove extra space before In addition

README.md
Huachao: also add firstByte description?

tests/test-timing.js
Huachao: > or >= ?

request.js
```diff
@@ -28,6 +28,7 @@ var http = require('http')
   , Multipart = require('./lib/multipart').Multipart
   , Redirect = require('./lib/redirect').Redirect
   , Tunnel = require('./lib/tunnel').Tunnel
+  , now = require('performance-now')
```
Huachao: do we must use the high resolution watch?

Huachao:
@nicjansma any updates, I think this feature is very important and necessary for request package

whiteboardmonk:
+1 for this feature

request.js
whiteboardmonk: I was looking at cURL results for quite a few API endpoints and I see it takes a few `ms` (10 - 50ms) on an average for most events. Although, a high-resolution watch would be nicer; as @mikeal suggested - we can possibly skip in interest of keeping the dependencies low.

request.js
```diff
@@ -749,6 +754,9 @@ Request.prototype.start = function () {
 
   if (self.timing) {
     self.startTime = new Date().getTime()
```
whiteboardmonk: Would recommend move to `Date.now()` instead of 'new Date().getTime()` as the former is twice as fast. Performance test: https://jsperf.com/date-now-vs-new-date

nicjansma:
Awesome, I'll work on all suggested changes shortly.

nicjansma:
Thanks everyone for their feedback, I've pushed a new change.

I've removed the `performance-now` package (which was just a lightweight wrapper over `window.performance.now()` in the browser and `process.hrtime()` in Node).  Note that I'm still using `process.hrtime()` / `performance.now()` (and added two small functions to facilitate this) instead of `Date.now()`.  `Date.now()` really shouldn't be used for anything that involves measuring timestamp deltas, as it is not monotonically non-decreasing.  In other words, if your system clock changes, `Date.now()` reflects that, and you could see negative deltas between two consecutive timestamps.

[`process.hrtime()`](https://nodejs.org/api/process.html) and [`window.performance.now()`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) were built to address this by guaranteeing that they're not sync'd to the system clock, so will give you the "actual" relative time regardless of how the system clock changes.  They also have the small benefit of giving you higher resolution (microseconds instead of milliseconds, though we're not exposing microsecond times).

Using `process.hrtime()` or `performance.now()` is relatively simple so if we don't want to include the `performance-now` package we just need one state variable and two functions to do the same thing.

I've also fixed the documentation and other minor things.

Hope this helps!

README.md
```diff
@@ -812,7 +812,18 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
 
 ---
 
-- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution, and the result provided on the response's `elapsedTime` property. The `responseStartTime` property is also available to indicate the timestamp when the response begins.
+- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution, and the result provided on the response's `elapsedTime` property. The `responseStartTime` property is also available to indicate the timestamp when the response begins. In addition, there is a `.timings` object available with the following properties:
+  - `start`: Timestamp when `request()` was initialized
```
Huachao: Is the timestamp like `start`, `end` which is got from process.hrtime a useful time for user, since the result from process.hrtime is relative to an arbitrary time in the past, and not related to the time of day

README.md
```diff
@@ -812,7 +812,18 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
...
+  - `socket` Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires.  This happens when the socket is assigned to the request (after DNS has been resolved).
```
nicjansma: Fixed -- there were 15 other occurrences of double-spaces, so if you're going for consistency I've fixed those as well.

README.md
Huachao: @nicjansma if I want to know the `start` time of the request in Unix format, how can we achieve this?

nicjansma: That's available today as `request.startTime`

Huachao: sorry, I mean the start time of response, can we use responseStartTime, it is still available?

nicjansma: Yup!  I didn't break any of the additional timestamps, so `responseStartTime` is still available.

README.md
```diff
@@ -812,7 +812,18 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
 
 ---
 
-- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution, and the result provided on the response's `elapsedTime` property. The `responseStartTime` property is also available to indicate the timestamp when the response begins.
+- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution, and the result provided on the response's `elapsedTime` property. The `responseStartTime` property is also available to indicate the timestamp when the response begins. In addition, there is a `.timings` object available with the following properties:
+  - `start`: Timestamp when `request()` was initialized
+  - `socket` Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires. This happens when the socket is assigned to the request (after DNS has been resolved).
+  -  `connect`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_connect) module's `connect` event fires. This happens when the server acknowledges the TCP connection.
```
Huachao: is there an extra space before `connect`

nicjansma: Good catch, fixed

nicjansma:
Any more concerns here?

Huachao:
@nicjansma I think the code is quite well, and @mikeal do you have any suggestions, I expect to see this feature very much 😃

Huachao:
@mikeal @simov , any concerns about the PR, when will this be merged, it's a long waiting feature

nicjansma:
@mikeal @simov Would love to see this merged in :)

nicjansma:
@simov Any concerns with this PR?

Huachao:
@FredKSchott please help to review this, thanks in advance.

FredKSchott:
I've only been going through the small, low-hanging PRs. But I appreciate the time and effort you and @nicjansma have given to writing this and continuing to ping/bump it. I'll try to give it a look tonight/tomorrow.

README.md
```diff
@@ -287,13 +287,13 @@ The method form takes parameters
 `auth(username, password, sendImmediately, bearer)`.
 
 `sendImmediately` defaults to `true`, which causes a basic or bearer
-authentication header to be sent.  If `sendImmediately` is `false`, then
```
FredKSchott: In the future, address these unrelated changes in a separate PR. It makes the review smaller/easier.

README.md
FredKSchott: +1 glad you thought about the deprecated behavior.

Can you add a comment near wherever `elapsedTime` and `responseStartTime ` are set in the code to indicate that they are now deprecated in favor of the `.timings` property? Once this has had some time to bake we can remove any README references to the old properties and think about how we deprecate them.

tests/test-timing.js
```diff
@@ -31,8 +31,19 @@ tape('non-redirected request is timed', function(t) {
     t.equal(err, null)
     t.equal(typeof res.elapsedTime, 'number')
     t.equal(typeof res.responseStartTime, 'number')
+    t.equal(typeof res.timings, 'object')
     t.equal((res.elapsedTime > 0), true)
     t.equal((res.responseStartTime > r.startTime), true)
+    t.equal((res.timings.start > 0), true)
```
FredKSchott: Do a `t.deepEqual(res.timings, ...)` here instead, so that we check that only the properties we expect are present.

FredKSchott:
re: [`performance-now`](https://www.npmjs.com/package/performance-now): I checked the library and it seems well tested, well supported, and has zero dependencies. I know that we're trying to remove dependencies and not add them, but by using `performance-now` we get to remove our custom implementation that is less tested, has never been used in production, and that leaves us responsible for all bugs. To me that sounds like a no brainer, as long as it continues to have a small footprint.

Not sure if @mikeal would still disagree, and I know that you already removed it. But if you still think it's the right call I'd say add it back (in a separate commit that can be easily reverted) and I'll give everyone a chance to weigh in before making the final call to merge.

README.md
nicjansma: No problem -- it was requested that my new documentation remove double-spaces for consistency, so I wanted to fix other occurrences.

tests/test-timing.js
nicjansma: Yeah, that's why I hadn't used `.deepEqual` initially.  I've added a check to get all property names and compare.

nicjansma:
@FredKSchott Deprecation notes added in https://github.com/request/request/pull/2452/commits/d9e8c3d5b4ccb98cb9c51d1775bd8c668f51e655 and tests updated in https://github.com/request/request/pull/2452/commits/a9ad38a3f7c4bc388af8af353b54fc05cfb81f63.

I'd certainly prefer to use `performance-now` for all of the reasons you suggested.  I've re-added it in https://github.com/request/request/pull/2452/commits/3f57975212976c88fc006e9e7b3e2cb5b53a2ddf, if everyone agrees.

Thanks for the feedback!  Let me know if there's anything else.

tests/test-timing.js
FredKSchott: Awesome 👍

README.md
FredKSchott: Thanks for the context, hard to get caught up when Github minimizes everything that was previously said.

FredKSchott:
LGTM, merging now

nicjansma:
Thanks!

mcorb:
@nicjansma This is great. I have a quality-of-impementation issue after trying to consume the new request timing API yesterday. Please consider:

We found it challenging to separate timeline entries (start, socket, connect, response, end) from time duration entries (dns, tcp, firstByte, download, total).

Because deltas and times are currently mixed together in the 'timings' object, users of the request library have to separate them manually before use.

This means each user currently has to hard-code the key names into their code in order to present the data to the user, chart it or do whatever they want.

A cleaner way to implement this API would be to only provide the absolute values (start, socket, connect, response, end) in your 'timeline' object.

You could then provide a convenience accessor containing 'timings' or deltas (dns, tcp, firstByte, download, total) computed from the absolute times in 'timeline'.

That way users will be able to just pass response.timeline or response.timings into their database, charting library, UI or whatever without processing, based on the one they need. And it future-proofs your interface so you can add new measurements (e.g. 'ssl') to request in future without breaking apps.

mcorb:
Another QoI issue we had is that the timestamps leak the process start time. This is just an implementing detail of performance-now that shouldn't be leaked into the request API. It'll be best to present either zero-based or session-based timings that can be piped straight into a metrics tool or processed more easily by developers.

I'm going to prepare two PRs to address both points today or tomorrow - @FredKSchott if you can hold any release until we've had a change to look at this that'd be great.

mcorb:
We found a lot of issues and decided to hook the core http agent module for our need.

This means I won't have the time to submit changesets but I'll try to help if you have any questions.

Some notes from a run against our in-hous http timing test suite:

1. We found `performance-now` to be broken. They introduce two rounding errors by going through nanoseconds and back, and also by not using the documented base time feature correctly. You might want to investigate using `process.hrtime()` directly.
2. You're not timing DNS correctly. Need to hook the socket 'lookup' event.
3. The results are incorrect for common use modes like HTTP keep-alive, connection reuse and connection pools.
4. More clarity needed on http redirect handling.

To get this right I recommend running against an HTTP timing test suite.

nicjansma:
Hi Mike!

Lots of great feedback.  Let me try to unpack some of it.

* RE: separating the timestamps (`start`, `socket`, `connect`, `response`, `end`) from the deltas (`dns`, `tcp`, `firstByte`, `download`, `total`).  I don't have strong feelings about this, but I can see how it would be helpful to have them separated.   I might suggest keeping `.timing` for the timestamps and `.timingDeltas` for the deltas.

    Does anyone else have a preference here?

* RE: timestamps leak the process start time.  I'm not sure I fully follow your concern.  It's true that all of the timestamps such as `.start`, `.socket`, etc are using the process' start time as the epoch (0-base).  This allows you to, for example, lay out all of the requests within your app in a timeline of requests, and each `.timings` timestamp can be directly placed on that timeline.

    Are you suggesting, instead, that `.start` process start epoch, while the other timestamps are using "time-since-`.start`" as epoch?  If so, I think that would be more confusing to have timestamps with different epochs.

    FWIW, [ResourceTiming](https://www.w3.org/TR/resource-timing-1/) keeps everything in a [DOMHighResTimeStamp](https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp), whose epoch is the start of the navigation.  `.startTime`, `.domainLookupStart`, etc are all using the same 0-base of navigation start (i.e. process start).

    Anyone that needs timestamps in a different epoch (e.g. `.socket` is the time since `.start`), could easily do the math, right?

* RE: 1. `performance-now` concerns.  I'm not seeing the rounding error, can you share details?  (My personal feeling is that most people won't look much beyond milliseconds at best -- I personally wouldn't be concerned about nanoseconds being rounded).

    As far as using the [`process.hrtime()` third parameter](https://github.com/nodejs/node/blob/ee7b7f693559afa05ea42279f856b4fc2186f527/lib/internal/process.js#L87) to automatically calculate the delta, I couldn't see how using it would give additional accuracy benefits vs. doing the same math?

    Anyways, I would suggest opening these Issues against [`performance-now`](https://github.com/braveg1rl/performance-now) if you think it could be improved.

* RE: 2. DNS.  Thanks for finding that!  I hadn't found the `lookup` event.  Fixes in: https://github.com/request/request/pull/2566

* RE: 3. Incorrect results for HTTP keep-alive, connection re-use and connection pools.

    Great catch.  I've tested a bit more with `forever` (for Keep Alive), `agent` (connection reuse) and `pool` (connection pool), and I believe have fixed the timestamps and deltas.  Fixes in https://github.com/request/request/pull/2566

* RE: 4. HTTP redirect handling.  The timings will report the data for the _final hop_ in a redirect chain.  I've improved the documentation to address this: https://github.com/request/request/pull/2566

    If someone wants extra credit, they could expose a log of the timings for all redirects. :)

mcorb:
Neat Nic, that was quick. I haven't had a chance to look at your patch but here are some quick replies to your points:

Start of Navigation is meaningful for a browser page load because all the timed loads are initiated by the navigation and immediately follow it.

In contrast, node's process start time isn't related to the use of the request() API. Node processes run for weeks or months and outbound http queries are triggered by timers, user interaction or all kinds of other actions any time. It's best not to bake that assumption into request()

I see two ways to go:
1. Just keep it dead simple and provide all absolute times as `Date.now()`
2. Include a start date in the parent object with `Date.now()` and make all the sub-timings relative to that.

All http libraries I know do one of the above, with my mild preference being (2). These are both universal and well-known formats, while process-relative time isn't.

Process-relative time is mostly used for performance instrumentation of a single process (compiler debugging, code optimisation) so it doesn't feel right here. Maybe that's the intended audience for `performance-now`?

Overall `Date.now()` is perfectly fine for this use case. You're measuring events based on the node event loop which already have a lower resolution, so `process.hrtime()` and `process-now` are just measuring noise using an API that was designed to measure CPU performance in nanoseconds in a single tick.

As for the flaw in `process-now`: the code is currently converting `process.hrtime()` to JS before doing its calculation so the nanoseconds are truncated, defeating the point of using high-resolution timers. I'll send them a patch but honestly `Date.now()` seems more appropriate for this use case.

mcorb:
By the way Nic, I totally respect your point of view and your work in this space. Some of this boils down to differences in developer expectations coming from client-side timing vs server-side/node conventions.

As long as the deltas are stored separately to absolute timings the API is future-proof and can be worked on incrementally.

(For completeness, there are other options too - like emitting an event or calling a user-supplied callback for each phase change so that they can decide how they measure, and also allowing for a real-time GUI like the Chrome network inspector instead of waiting till the end of the request. You might want to move your collections into a private `markTime()` function for clarity.)

By the way, did you test that your patch supplies timings for *failed* requests? DNS and connection timings are equally, if not more important for failures too. I didn't check if your revision addresses this - it was in my notes from my previous review that I forgot to mention.

Sorry to come at you with all this review. I'm not a direct consumer of this API any more so please pick what you find helpful and go with your best judgement.

nicjansma:
Again great feedback Mike, appreciate your thoughts.

Unfortunately there are several reasons to avoid `Date.now()` if possible when doing performance measurements (please see my [previous comment](https://github.com/request/request/pull/2452#issuecomment-264708292) for details).

But if we only use `Date.now()` once per request -- to log the start "epoch" wall time -- and use `process.hrtime()` for all other timestamps, we gain the advantage of a higher-resolution, monotonically non-decreasing clock to measure all of the phases.

I like your idea for (2).  I've made some changes in the other PR https://github.com/request/request/pull/2566 for how the data is presented based on your recommendations:

1. There is a new `timingStart` property that is based on `Date.now()`
2. The `timings` timestamps are relative to `timingStart`, using `process.hrtime()`.  This object now only contains `socket`, `lookup`, `connect`, `response` and `end` (moving the durations out).  `start` is no longer needed, since it will always in effect be `0`.
3. The other durations were moved to `timingPhases`.

I guess I still don't understand your concern about `performance-now` -- looking forward to your patch :)  (Remeber that `Date.now()` is limited to millisecond resolution, so truncating nanoseconds is still better accuracy than `Date.now()`).

For the DNS and TCP failure cases you mention, no `response` object is created, so there's (currently) no way to get timings out of failures.

If you have any further feedback, want to move it to https://github.com/request/request/pull/2566?

Thanks!

Issue #2338, opened 2016-08-21
Update "hawk" in order to fix "boom" unmet dependency warning


mickaeltr:
Hello,

When installing "request", I get the following warnings:

```
npm WARN unmet dependency …/node_modules/request/node_modules/hawk requires boom@'2.x.x' but will load
npm WARN unmet dependency …/node_modules/boom,
npm WARN unmet dependency which is version 3.2.2

npm WARN unmet dependency …/node_modules/request/node_modules/hawk/node_modules/cryptiles requires boom@'2.x.x' but will load
npm WARN unmet dependency …/node_modules/boom,
npm WARN unmet dependency which is version 3.2.2
```

`request` currently depends on `hawk ~3.1.3`. It seems that updating to `hawk ~5` would fix these warnings.

Thanks

wtgtybhertgeghgtwtg:
It seems that `hawk ~5` requires at least node v4.  Wouldn't that mean that `request` would have to bump the minimum required node version to v4 for something like this to happen?

JSteunou:
should be closed

mickaeltr:
I am closing it as it is outdated.

JSteunou:
@mickaeltr : I was searching for all "Hawk" related issues, 99% of them can be closed :) I started by commenting this one but then was too lazy to continue :D

Issue #2436, opened 2016-10-25
Hawk Dependency Vulnerability - Regular Expression Denial of Service 


krzyszt:
Is there  a chance to update hawk dependency to version 4.1.1 or greater?

Specifically crafted long headers or uris can cause a minor denial of service when using hawk versions less than 4.1.1. as per

https://nodesecurity.io/advisories/77

simov:
3.1.3 that we are using is patched:

```
Patched: >=3.1.3 < 4.0.0 || >=4.1.1
```

wtgtybhertgeghgtwtg:
Now that `request` [only supports `node >= 4`](https://github.com/request/request/commit/01aefdd87cdd9cc44af6828c55fe5ef9c7e205d7), can `hawk` be updated?

sahas-:
Can we get this upgraded pls?

<img width="670" alt="screen shot 2017-01-18 at 8 27 33 pm" src="https://cloud.githubusercontent.com/assets/10361685/22091517/c764669a-ddbc-11e6-9e19-0f1f3d99db53.png">

PR #2456, opened 2016-11-08
Hawk update to fix node security issues.

Abhinay077:


joepie91:
Assuming that this PR refers to [this security advisory](https://nodesecurity.io/advisories/77), it would not be necessary - the issue was resolved in version 3.1.3, which is the minimum version that `request` already requires.

Or is this meant to fix a different vulnerability?

FredKSchott:
This should be okay to merge now that we've dropped 0.x support, any objections?
Fixes https://github.com/request/request/issues/2338 & https://github.com/request/request/issues/2436.

ashawley:
Fixed in #2751 and released in 2.82.0.

PR #2460, opened 2016-11-12
Fix wrong MIME type in example

OwnageIsMagic:


FredKSchott:
Had to look this up since I didn't know off the top of my head, but the change is correct. See: http://stackoverflow.com/questions/33692835/is-the-mime-type-image-jpg-the-same-as-image-jpeg

OwnageIsMagic:
Thank you
The same problem here 
https://github.com/form-data/form-data/pull/285
It took me hour to reveal why server rejects my images

FredKSchott:
happy to help. I don't have merge permissions to form-data but I commented

PR #2463, opened 2016-11-17
Fixing requests/requests issue #2462 - AWS support does not include t…

### LINKED ISSUE #2462: AWS support does not include the use of session tokens for temporary credentials
The AWS option currently only supports cases wherein a permanent access/secret key is supplied as arguments, not for temporary keys (such as those retrieved via STS). This is a common use case, wherein a user can authenticate via something other than having their own IAM user, and can then be given credentialed access to various AWS endpoints (for instance, access to an API in the API Gateway).

Fixing this is extremely simple (though there's a limitation in the aws4 library in certain cases that requires knowledge of AWS, see https://github.com/mhart/aws4/issues/30 for details), and I'll be submitting a PR for it.

lostcolony:
Modified readme to indicate support of 'session' as part of the AWS object.

Modified call to aws4.sign to include session token if provided.

Set x-amz-security-token header if session was included.

Modified tests appropriately to ensure x-amz-security-token is in place if session is provided, and is still excluded if session is not.

Fixes #2462


simov:
:+1:

PR #2469, opened 2016-11-21
Catch piped stream errors and emit them properly.

SzybkiSasza:
Added proper handler for errors caused by piped streams. Without it, streaming multipart data could  break the lib, if provided stream throws an error.

### DIFF
--- a/request.js
+++ b/request.js
@@ -516,6 +516,9 @@ Request.prototype.init = function (options) {
       }
     }
 
+    // Allow for catching stream errors in parent context.
+    src.on('error', function() {})
+
     // self.on('pipe', function () {
     //   console.error('You have already piped to this stream. Pipeing twice is likely to break the request.')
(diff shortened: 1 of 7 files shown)


SzybkiSasza:
What about this one? Does anyone care? The PR contains proper tests and fix...

FredKSchott:
Thank you for your contribution.

Unfortunately this isn't expected behavior for streams. If an error occurs on the source stream piped into `request`, it is expected that it is emitted from the source stream and not propagated and re-emitted down the chain. I personally hate this, but propagating source errors by default would break expected stream behavior. See http://grokbase.com/t/gg/nodejs/12bwd4zm4x/should-stream-pipe-forward-errors for a deeper explanation. 

You can use a library like https://github.com/mafintosh/pump with request to get the behavior you're desiring here.

PR #2480, opened 2016-12-03
dependency(har-validator): update to ~v4.2.0

ahmadnassri:
- `v3.x` introduces massive performance improvement.
- `v4.x` minimizes dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "extend": "~3.0.0",
     "forever-agent": "~0.6.1",
     "form-data": "~2.1.1",
-    "har-validator": "~3.3.0",
+    "har-validator": "~4.1.0",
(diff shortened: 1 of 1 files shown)


simov:
Probably version 4.x will be even better regarding the dependencies count :) I'm glad you removed the CLI related modules from the core har module.

ahmadnassri:
indeed. as soon as I made this PR, I quickly followed with more cleanup! I shall update it now to 4.x

ahmadnassri:
PR updated.

ahmadnassri:
struggled getting the local setup to test needed browser fixes for the test here to succeed, seems to be only triggering on Travis ... but finally got it.

PR #2514, opened 2017-01-16
Change tags to keywords in package.json

humphd:
I'm in the process of teaching my students about how to do PRs, and trying @mafintosh's [first PR idea](https://twitter.com/mafintosh/status/782288714785492992) with them.  As I was writing up an example for their lab, I noticed that the `package.json` here uses `tags` vs. `keywords.

This fixes it to properly include the existing keywords, which aren't currently being picked up by npmjs, see:

screen shot 2017-01-16 at 12 19 07 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "request",
   "description": "Simplified HTTP request client.",
-  "tags": [
+  "keywords": [
     "http",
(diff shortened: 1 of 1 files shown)


FredKSchott:
+1, good catch!

Issue #2546, opened 2017-02-16
A couple of 'issues' on digest authentication implementation


albertsuchvi:
Hi,

more than real issues these are potential improvements to current implementation of digest authentication...

Background: we have a server that uses digest auth because, due to some contraints on the network connection between server and client, we cannot fully protect for MitM attacks, and consequently we cannot use basic auth.

As far as I understand, current implementation in request is that, when setting the immediately parameter to false, it first sends a non authenticated request and then, when the server requests digest, it sends the digest authenticated request. The problem with this implementation is that a MitM may intercept the initial request and sets authentication required to basic rather than digest, effectively 'downgrading' teh authentication and intercepting the credentials.

A potential solution would be controlling through an additional field in the auth parameter object which authentication mechanisms are allowed in the second request, so basic authentication can be fully disabled.

There is also a potential improvement: currently for every request that wants to use digest auth, two http requests must be sent: one non authenticated and then the one authenticated with digest. It would be possible to keep the parameters from the previous digest authentication (including new nonce,...) and try to use them to authenticate directly for succesive requests.

I am looking at the code, and I think that both improvements are not difficult to implement. If you think they make sense, I can make a pull request once I have the changes ready

tx

albertsuchvi:
Hi,

any feedback on this topics?. I have done the changes and I am ready to PR them. 
I will wait for a couple of days for feedback before doing the pull request

PR #2547, opened 2017-02-17
Improvements in digest authentication support

albertsuchvi:
As I mentioned in issue #2546, there are a couple of improvements in the implementation of digest authentication:

1.- have a mechanism to make sure that digest authentication is used:
What I did is adding a new (optional) parameter to the auth() method that can disable the use of specific authorization methods when sendImmediately is set to false. With this parameter, teh caller can disable, for example, basic and bearer authentication, thus ensuring that only digest authentication is used. 

2.- optimize the use of digest authentication when consecutive requests are issued
I added a new class variable to Auth to store the information from the previous sent digest authentication. When a new request with sendImmediately=false is sent, and if there was a previous digest authentication sent, it tries to send dircetly a new digest auithentication using the parameters (qop, nonce, cnonce, seq,...) from the revious one. If the request does not work due to baud authentication, the currently implemented mechanism if resending the request with teh authentication required by the server will work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,15 +282,22 @@ If passed as an option, `auth` should be a hash containing values:
 - `pass` || `password`
 - `sendImmediately` (optional)
 - `bearer` (optional)
+- `disable` (optional)
 
 The method form takes parameters
-`auth(username, password, sendImmediately, bearer)`.
+`auth(username, password, sendImmediately, bearer, disable)`.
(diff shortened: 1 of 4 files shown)


albertsuchvi:
I missed one test not passing and a minor problem in lint. Will fix them and pull request again....

PR #2551, opened 2017-02-19
Add host and port to timed out error properties

ruimarinho:
Allows for easier debugging when a timeout occurs due to a timed out error. Based on https://github.com/request/request/pull/1747.


request.js
```diff
@@ -778,6 +778,8 @@ Request.prototype.start = function () {
           var e = new Error('ESOCKETTIMEDOUT')
           e.code = 'ESOCKETTIMEDOUT'
           e.connect = false
+          e.host = self.host
+          e.port = self.port
```
ruimarinho: @FredKSchott any idea why `self.port` is kept as string after url parse? I had to cast the test to string.

FredKSchott: That is bizarre, it looks like it's a number everywhere else. See https://github.com/seegno-forks/request/blob/88f9a04572c2fec21bf46ab2628950c3fae8bf20/request.js#L303-L314

FredKSchott: Ah no, url.parse returns port as a string. But we're setting it manually as a number in other places . 

We should change to always be a number or string in a future PR.

tests/test-timeout.js
```diff
@@ -65,6 +65,20 @@ tape('should set connect to false', function(t) {
...
+tape('should set host and port', function(t) {
+  var shouldTimeout = {
+    url: s.url + '/timeout',
+    timeout: 100
+  }
+
+  request(shouldTimeout, function(err, res, body) {
+    checkErrCode(t, err)
+    t.ok(err.host === 'localhost', 'Read Timeout Error should set \'host\' property')
+    t.ok(err.port === String(s.port), 'Read Timeout Error should set \'port\' property')
```
FredKSchott: use the actual value of s.port here (`'61160'`), similar to how you did on L76. That will make the test more accurate and solve your casting problem.

request.js
ruimarinho: Yep, I haven't spent much time reading the code but it may have introduced a bug already.

tests/test-timeout.js
ruimarinho: The port is dynamic so that's not possible :/

franciscocardoso:
Hi @FredKSchott,

Is there any plans to merge this PR in the near future?

FredKSchott:
@franciscocardoso just rebased. If travis passes, I'm good to merge this.

FredKSchott:
Huh, looks like tests aren't passing anymore. @ruimarinho if you can get tests passing again I'm 👍to merge

ruimarinho:
This is sitting at an unsupported fork right now. Will open a new PR.

PR #2560, opened 2017-02-23
Replace lib/getProxyFromURI.js with proxy-from-env

Rob--W:
## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A

## PR Description
This is identical to PR #2088 (see the detailed description over there), except rebased on master. Somehow the previous PR [cannot be re-opened](https://github.com/request/request/pull/2088#issuecomment-281810732), so here is a new PR.

/cc @FredKSchott


tests/test-proxy.js
```diff
@@ -138,13 +138,13 @@ function addTests() {
       url    : 'https://google.com',
       tunnel : false,
       pool   : false
-    }, true)
+    }, false)
```
FredKSchott: why are we no longer using a proxy for these requests? Shouldn't this change require no change to testing?

FredKSchott:
Please update the README to document the new support for ALL_PROXY (and maybe link to https://github.com/Rob--W/proxy-from-env#environment-variables for more information.)

We're trying to prune down our number of dependencies, but since this has no dependencies of its own, seems more reliable and better tested than our implementation, and reduces the amount of code we need to maintain... this seems like a reasonable improvement.

tests/test-proxy.js
Rob--W: Because `http_proxy` is only supposed to affect http-requests, not https. Use `all_proxy` to get the original behavior.

Rob--W:
I have edited the version in the first commit (bump to `proxy-from-env@1.0.0`) and added a separate commit for the README change. Let me know if I should squash the commits.

The change may technically be a breaking change, because previously `http_proxy` was also used for `https_proxy` (and now it is not any more). request's README did not mention this behavior (and the de-facto standard is to not use `http_proxy` as a fallback to `https_proxy`), so hopefully nobody relied on it.

Note that I did not introduce new unit tests with this PR because the proxy-from-env package already has an extensive sets of unit tests.

tests/test-proxy.js
FredKSchott: Got it.

README.md
```diff
@@ -546,9 +546,11 @@ The following environment variables are respected by `request`:
 
  * `HTTP_PROXY` / `http_proxy`
  * `HTTPS_PROXY` / `https_proxy`
+ * `ALL_PROXY` / `all_proxy`
  * `NO_PROXY` / `no_proxy`
 
 When `HTTP_PROXY` / `http_proxy` are set, they will be used to proxy non-SSL requests that do not have an explicit `proxy` configuration option present. Similarly, `HTTPS_PROXY` / `https_proxy` will be respected for SSL requests that do not have an explicit `proxy` configuration option. It is valid to define a proxy in one of the environment variables, but then override it for a specific request, using the `proxy` configuration option. Furthermore, the `proxy` configuration option can be explicitly set to false / null to opt out of proxying altogether for that request.
+If neither `http_proxy` nor `https_proxy` is set, then `all_proxy` is used if available.
```
FredKSchott: Describe what happens if none of the three are set/available

FredKSchott:
One small comment, but otherwise lgtm.

Thanks for reopening and working with us to get this merged.

FredKSchott:
Can you provide some links/docs on "and the de-facto standard is to not use http_proxy as a fallback to https_proxy"? I'm seeing a lot of people who are doing what request is currently doing so I'm a little skeptical: https://github.com/search?utf8=%E2%9C%93&q=process.env.HTTP_PROXY+https&type=Code&ref=searchresults

The breaking change is my one concern with merging this in. While it wasn't documented, it was fairly easy to rely on and the amount that this would break for those users is significant. With ~1M downloads a day that's something we need to consider.

Would you support an option in `proxy-from-env` to use the HTTP fallback if HTTPS is not available? If we could use that I would feel comfortable merging.

README.md
Rob--W: Do you really see value in stating the obvious that if no proxy config is set, that the requests are not proxied?

Rob--W:
> Can you provide some links/docs on "and the de-facto standard is to not use http_proxy as a fallback to https_proxy"?

See https://github.com/Rob--W/proxy-from-env#external-resources. I consider curl and wget as canonical reference implementations.

> The breaking change is my one concern with merging this in. While it wasn't documented, it was fairly easy to rely on and the amount that this would break for those users is significant. With ~1M downloads a day that's something we need to consider.
>
> Would you support an option in proxy-from-env to use the HTTP fallback if HTTPS is not available? If we could use that I would feel comfortable merging.

How about this:

```diff
diff --git a/request.js b/request.js
index dca5427..a6c2f2f 100644
--- a/request.js
+++ b/request.js
@@ -277,6 +277,10 @@ Request.prototype.init = function (options) {
 
   if (!self.hasOwnProperty('proxy')) {
     self.proxy = getProxyForUrl(self.uri) || null
+    if (!self.proxy && self.uri.protocol === 'https:') {
+      // Fall back to http_proxy when https_proxy is not set, for backwards-compatibility.
+      self.proxy = getProxyForUrl(self.uri.href.replace('s', '')) || null
+    }
   }
 
   self.tunnel = self._tunnel.isEnabled()
diff --git a/tests/test-proxy.js b/tests/test-proxy.js
index fe93cde..dd5cefb 100644
--- a/tests/test-proxy.js
+++ b/tests/test-proxy.js
@@ -138,13 +138,13 @@ function addTests() {
       url    : 'https://google.com',
       tunnel : false,
       pool   : false
-    }, false)
+    }, true)
 
     runTest('http_proxy environment variable and https: url', {
       env    : { http_proxy : s.url },
       url    : 'https://google.com',
       tunnel : false
-    }, false)
+    }, true)
 
     runTest('HTTPS_PROXY environment variable and https: url', {
       env    : { HTTPS_PROXY : s.url },
```

Rob--W:
I understand the desire for backwards-compatibility, but do note that automatically using `http_proxy` as a fallback implies that it is not possible to only proxy http requests and allow https requests to go through with a direct request.

A HTTP proxy is not necessarily capable of proxying HTTPS. The former looks similar to a regular HTTP request (the request is visible in plain text to the proxy server), whereas the latter requires a tunnel (usually with a CONNECT request) (the original request is encrypted and can only be decrypted by the operator of the website who shared the certificate).

Are you okay with showing a deprecation warning when the fallback is used?

FredKSchott:
@Rob--W sorry for the delay, the last two weeks have been too busy and I hadn't been able to find time to respond.

I completely agree that what you are proposing is "more correct" than request's current behavior (thanks for the curl/wget examples to confirm). But regardless, our current behavior is out there, and lots of people are now relying on it. To break those user's libraries/applications and tell them it was their fault would be doing a disservice to them.

I am a -1 on merging anything that causes a potentially serious breaking change to request. If your library won't be able to support the fallback, I would suggest we table this until it's time to for a v3.0 release.

Rob--W:
I'll update the PR (to resolve merge conflicts) once this is approved.

PR #2564, opened 2017-02-25
Update README.md warning on initialization errors

kamagatos:
Initialization errors should be caught within the callback. They won't emit the `error` event.

## PR Checklist:
       
### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,17 @@ request
   })
   .pipe(fs.createWriteStream('doodle.png'))
 ```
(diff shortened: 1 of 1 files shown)


FredKSchott:
@kamagatos thanks for the contribution!

Hmm, this seems like a bug more than a blessed way of doing things. I would rather we fix the issue instead of documenting a temporary workaround. @kamagatos Are you interested in creating a fix for this?

PR #2566, opened 2017-02-28
Timings: Tracks 'lookup', adds 'wait' time, fixes connection re-use

nicjansma:
## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: Addresses some feedback from https://github.com/request/request/pull/2452

## PR Description
Feedback from @mcorb in https://github.com/request/request/pull/2452 found a few items with the new `.timings` object.

Changes:

1. The [`net` `'lookup'`](https://nodejs.org/api/net.html#net_event_lookup) event is now tracked, which is the correct event to track DNS lookup time.
2. There is a new `.timingStart` property that is based on `Date.now()`.
3. The `.timings.*` timestamps are relative to `.timingStart` (using `now()`).
4. The `.timings` object now only contains `socket`, `lookup`, `connect`, `response` and `end` (moving the durations to `.timingPhases`)
5. `.timings.start` is no longer needed, since it will always in effect be `0`.
6. The other durations such as `wait`, `dns`, `tcp`, etc were moved to `.timingPhases`.
7. A new `.timingPhases.wait` delta is added, which tracks the time from `start->socket`
8. The `.timingPhases.dns` calculation is changed from `start->socket` to `socket->lookup`.
9. Documentation was updated to be clearer about the handling of HTTP redirects and for all of the new properties above.
10. For cases of connection re-use, the timings would have resulted in `NaN`s or negative numbers.  This is resolved by assuming any events (such as `lookup`) that did not happen (due to re-use) are "filled in" using the timestamp of the previous phase (such as `socket`), resulting in a `0` ms delta.
11. Fixes https://github.com/request/request/issues/2575 by not re-adding event listeners to Keep-Alive sockets (resulting in EventEmitter leak)

Here is an example of the current properties:

```
{ 
  elapsedTime: 143,                 // deprecated
  responseStartTime: 1488424737161, // deprecated
  timingStart: 1488424737018,       // Unix Epoch ms
  timings: {                        // relative to timingStart
     socket: 6.410719000000002,
     lookup: 7.299280000000001,
     connect: 46.287758000000004,
     response: 105.45501300000001,
     end: 142.975414 
  },
  timingPhases: {                   // durations
     wait: 6.410719000000002,
     dns: 0.8885609999999993,
     tcp: 38.988478,
     firstByte: 59.167255000000004,
     download: 37.52040099999999,
     total: 142.975414
  } 
}
```

Tests have been added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -819,7 +819,7 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
   - `timings` Contains event timestamps in millisecond resolution relative to `timingStart`. If there were redirects, the properties reflect the timings of the final request in the redirect chain:
     - `socket` Relative timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires. This happens when the socket is assigned to the request.
     - `lookup` Relative timestamp when the [`net`](https://nodejs.org/api/net.html#net_event_lookup) module's `lookup` event fires. This happens when the DNS has been resolved.
-    - `connect`: Relative timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_connect) module's `connect` event fires. This happens when the server acknowledges the TCP connection.
+    - `connect`: Relative timestamp when the [`net`](https://nodejs.org/api/net.html#net_event_connect) module's `connect` event fires. This happens when the server acknowledges the TCP connection.
     - `response`: Relative timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_response) module's `response` event fires. This happens when the first bytes are received from the server.
     - `end`: Relative timestamp when the last bytes of the response are received.
   - `timingPhases` Contains the durations of each request phase. If there were redirects, the properties reflect the timings of the final request in the redirect chain:
--- a/README.md
+++ b/README.md
@@ -812,17 +812,23 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
 
 ---
 
-- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution, and the result provided on the response's `elapsedTime` property. The `responseStartTime` property is also available to indicate the timestamp when the response begins. In addition, there is a `.timings` object available with the following properties:
-  - `start`: Timestamp when `request()` was initialized
-  - `socket` Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires. This happens when the socket is assigned to the request (after DNS has been resolved).
-  - `connect`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_connect) module's `connect` event fires. This happens when the server acknowledges the TCP connection.
-  - `response`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_response) module's `response` event fires. This happens when the first bytes are received from the server.
-  - `end`: Timestamp when the last bytes of the response are received.
-  - `dns`: Duration of DNS lookup (`timings.socket` - `timings.start`)
-  - `tcp`: Duration of TCP connection (`timings.connect` - `timings.socket`)
-  - `firstByte`: Duration of HTTP server response (`timings.response` - `timings.connect`)
-  - `download`: Duration of HTTP download (`timings.end` - `timings.response`)
-  - `total`: Duration entire HTTP round-trip (`timings.end` - `timings.start`)
+- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution. When set, the following properties are added to the response object:
+  - `elapsedTime` Duration of the entire request/response in milliseconds (*deprecated*).
+  - `responseStartTime` Timestamp when the response began (in Unix Epoch milliseconds) (*deprecated*).
+  - `timingStart` Timestamp of the start of the request (in Unix Epoch milliseconds).
+  - `timings` Contains event timestamps in millisecond resolution relative to `timingStart`. If there were redirects, the properties reflect the timings of the final request in the redirect chain:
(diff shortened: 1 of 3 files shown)


nicjansma:
@Huachao and @FredKSchott would appreciate you taking a look!

request.js
```diff
@@ -866,30 +877,50 @@ Request.prototype.onRequestResponse = function (response) {
...
+      if (!self.timings.socket) {
+        self.timings.socket = 0
+      }
+      if (!self.timings.lookup) {
+        self.timings.lookup = self.timings.socket
+      }
+      if (!self.timings.connect) {
+        self.timings.connect = self.timings.lookup
+      }
+      if (!self.timings.response) {
```
Huachao: Q: in which case `self.timings.response` is blank, or `response` event is not triggered?

README.md
```diff
@@ -812,17 +812,20 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
...
-  - `connect`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_connect) module's `connect` event fires. This happens when the server acknowledges the TCP connection.
-  - `response`: Timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_response) module's `response` event fires. This happens when the first bytes are received from the server.
-  - `end`: Timestamp when the last bytes of the response are received.
-  - `dns`: Duration of DNS lookup (`timings.socket` - `timings.start`)
+- `time` - If `true`, the request-response cycle (including all redirects) is timed at millisecond resolution, and the result provided on the response's `elapsedTime` property. The start time of the request (in Unix Epoch milliseconds via `Date`) is saved as `timingStart`. The `responseStartTime` property is also available to indicate the timestamp when the response begins. In addition, there is a `timings` object available with timestamps in millisecond resolution relative to `timingStart`. If there were redirects, the properties reflect the timings of the final request in the redirect chain:
```
Huachao: I think it's important to emphasize the `elapsedTime` and `responseStartTime ` is __deprecated__ in README

tests/test-timing.js
```diff
@@ -27,23 +27,31 @@ tape('setup', function(t) {
...
+    t.equal((res.timingStart >= start), true)
     t.equal(typeof res.timings, 'object')
     t.equal((res.elapsedTime > 0), true)
+    t.equal((res.elapsedTime <= (end - start)), true)
     t.equal((res.responseStartTime > r.startTime), true)
-    t.equal((res.timings.start > 0), true)
-    t.equal((res.timings.socket >= res.timings.start), true)
+    t.equal((res.timings.socket >= 0), true)
+    t.equal((res.timings.lookup >= res.timings.socket), true)
     t.equal((res.timings.connect >= res.timings.socket), true)
```
Huachao: res.timing.lookup

request.js
nicjansma: I'm not sure it's even possible, but I added it here for consistency and to protect against edge cases.

README.md
nicjansma: Good idea, I've reorganized the description here a bit to be a list of all properties added.

tests/test-timing.js
nicjansma: Great catch.

FredKSchott:
Thanks for the work @nicjansma, I'll be mostly offline until Monday but would love to take a look at this once back

README.md
```diff
@@ -812,17 +812,23 @@ default in Linux can be anywhere from 20-120 seconds][linux-timeout]).
...
+  - `elapsedTime` Duration of the entire request/response in milliseconds (*deprecated*).
+  - `responseStartTime` Timestamp when the response began (in Unix Epoch milliseconds) (*deprecated*).
+  - `timingStart` Timestamp of the start of the request (in Unix Epoch milliseconds).
+  - `timings` Contains event timestamps in millisecond resolution relative to `timingStart`. If there were redirects, the properties reflect the timings of the final request in the redirect chain:
+    - `socket` Relative timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires. This happens when the socket is assigned to the request.
+    - `lookup` Relative timestamp when the [`net`](https://nodejs.org/api/net.html#net_event_lookup) module's `lookup` event fires. This happens when the DNS has been resolved.
+    - `connect`: Relative timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_connect) module's `connect` event fires. This happens when the server acknowledges the TCP connection.
```
Huachao: the connect event is not the one in `http` module, it's the one in `net` of `Socket` class, the doc url should be `https://nodejs.org/api/net.html#net_event_connect`

nicjansma: Great catch, fixed

nicjansma:
@mikeal Some fixes in here for https://github.com/request/request/issues/2575

mikeal:
Once CI passes I would like to get this in because it fixes a leak introduced in the last release. Is anyone still in the middle of reviews that would object?

mikeal:
Wow, there's a big falloff in code coverage.

nicjansma:
-0.35% overall.  It seems like it's just due to 4 lines in the new patch that aren't run (which cover error conditions):

https://codecov.io/gh/request/request/compare/667e923d8d51a0f0fb4c66c91d94a40f3d3e575a...32232d91f86f16b20d76f21363d7ae4a22a0af45/diff

If you're worried about it, I can see if I can create test cases for those scenarios in a followup PR (though tomorrow at the earliest).

mikeal:
Oh i see, was reading that wrong. Merging this to fix the bug. Would love a followup with more tests :)

nicjansma:
@mikeal KeepAlive test here: https://github.com/request/request/pull/2590

PR #2571, opened 2017-03-04
Correctly format the Host header for IPv6 addresses

### LINKED ISSUE #2292: IPv6 address misinterpreted
there are already a couple of IPv6 related issues filed, #1705 #1944 but all are different with this new one:

my server is binding on IPv6 address `[::]`, means same as IPv4's `0.0.0.0`; in dual stack OSes (like Linux), this resulted being able to handle both IPv4 and IPv6 clients, but this request package is handling IPv6 address wrong, see this example:

I use nc to simulate an http server listening on IPv6 any address [::] port 3000; this listening address and port can be confirmed with `lsof`

``` console
➸ nc -l6 -p 3000
...

➸ lsof -nPp 26678        # from another shell, find out the pid of nc is 26678
...
nc      26678  user    3u  IPv6 391972      0t0     TCP *:3000 (LISTEN)
```

then make a curl call of this server, with IPv6 address, like `curl -v [::1]:3000/any-url` the nc side can echo what is actually sent in the request:

``` console
➸ nc -l6 -p 3000
GET /any-url HTTP/1.1
Host: [::1]:3000
User-Agent: curl/7.47.0
Accept: */*

```

The `Host: [::1]:3000` is sent by curl as a http client, which is correct;

however, with the request as the http client, this is actually sent on wire:

``` console
POST /any-url HTTP/1.1
host: ::1:3000
accept: application/json
content-type: application/json
content-length: 307
Connection: close

```

by request call code is

``` javascript
request.post(
    { url: 'http://[::1]:3000/any-url',
      json: { a: 2, ... } },
    (err, resp, body) => console.log(err, resp, body));
```

the `host: ::1:3000` part is wrong, IPv6 address should be enclosed in bracket.

BTW: I tested other nodejs http client as well, like this one http://restify.com/#jsonclient

``` console
➸ node
> var restify = require('restify');
undefined
> var client = restify.createJsonClient({ url: 'http://[::1]:3000/tasks' });
undefined
> client.post('/abc', {a: 2}, (err, req, res, obj) => console.log(err, req, res, obj));

```

then the nc server side got correct result:

``` console
POST /abc HTTP/1.1
accept: application/json
content-type: application/json
user-agent: restify/4.1.1 (x64-linux; v8/5.0.71.52; OpenSSL/1.0.2h) node/6.2.2
date: Wed, 20 Jul 2016 22:27:07 GMT
content-length: 7
content-md5: qrRX4OwkT0d+4MCXuUonKA==
Host: [::1]:3000
Connection: keep-alive

{"a":2}
```

JamesMGreene:
Fixes #2292

## PR Checklist:
       
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: [**N/A**, or add link to issue here]
       

## PR Description

When using IPv6 addresses in URLs, the address must be enclosed in brackets, e.g.`https://[::1]:3000/headers.json`

Likewise, when using IPv6 addresses in `Host` header values, the address again must be enclosed in brackets, e.g. `Host: [::1]:3000`

When the **request** module adds the `Host` header value, it does so using the `hostname` and `port` (if needed) parts of the parsed URL.  However, this approach does not correctly format IPv6 addresses with the necessary enclosing brackets as `hostname` does not include them.

If you were instead to utilize the `host` portion of the parsed URL instead, you would automatically get the correctly bracketed value, potentially plus the port.  My testing has not shown any case where the port was unnecessarily included in the `host` value so I have also avoided adding extra code to remove the port if deemed unnecessary by **request**, though I'd be happy to add such if anyone can prove it is needed.

### DIFF
--- a/request.js
+++ b/request.js
@@ -289,13 +289,10 @@ Request.prototype.init = function (options) {
   self.setHost = false
   if (!self.hasHeader('host')) {
     var hostHeaderName = self.originalHostHeaderName || 'host'
-    self.setHeader(hostHeaderName, self.uri.hostname)
-    if (self.uri.port) {
-      if ( !(self.uri.port === 80 && self.uri.protocol === 'http:') &&
-           !(self.uri.port === 443 && self.uri.protocol === 'https:') ) {
-        self.setHeader(hostHeaderName, self.getHeader('host') + (':' + self.uri.port) )
-      }
-    }
+    // When used with an IPv6 address, `host` will provide
+    // the correct bracketed format, unlike using `hostname` and
+    // optionally adding the `port` when necessary.
+    self.setHeader(hostHeaderName, self.uri.host)
     self.setHost = true
   }
 
--- a/tests/test-headers.js
+++ b/tests/test-headers.js
@@ -4,6 +4,7 @@ var server = require('./server')
   , request = require('../index')
   , util = require('util')
   , tape = require('tape')
+  , url = require('url')
 
 var s = server.createServer()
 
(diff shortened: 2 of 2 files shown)


mikeal:
Nice catch. request predates that field being correct in core's URL parser :+1:

JamesMGreene:
@mikeal: Any chance you know what versions of Node would **not** correctly format this due to the URL parser?

JamesMGreene:
FWIW, I tested this in `0.12.17` and it seems to work correctly.

P.S. Thanks for the quick response and merging. Any idea when a new release will be published?

mikeal:
Honestly, this was probably fixed like 5 years ago in core :)

Request was initially written to Node.js 0.1.x, it pre-dates npm :)

JamesMGreene:
I tested this against Node `0.10.48` just for the heck of it but it looked like there were some other issues with the tests against that version.

JamesMGreene:
That's awesome that this module pre-dates NPM, LOL. 😮 

One more time: any idea when a new release of "request" will be published?  We're dependent on this fix at work for an upcoming release in April.

mikeal:
v2.80.0

JamesMGreene:
Thanks so much! 💝

PR #2579, opened 2017-03-05
refactor(lint): replace eslint with standard

### LINKED ISSUE #2577: Move to standardjs code style
Look, it was cute to have my own code style for the first 6 years but this is just making it weird to contribute to now. I'm +1 for moving to standardjs. I'm -1 for moving to anything else, cause I just don't wanna have an argument about it :)

ahmadnassri:
## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: [#2577]

## PR Description
- [x] replace `eslint` with `standard`
- [x] make appropriate changes to style to pass a successful lint step


mikeal:
+1

I just went through doing all this myself and this is actually a bit better.

lib/oauth.js
```diff
@@ -32,11 +31,11 @@ OAuth.prototype.buildParams = function (_oauth, uri, method, query, form, qsLib)
     oa.oauth_signature_method = 'HMAC-SHA1'
   }
 
-  var consumer_secret_or_private_key = oa.oauth_consumer_secret || oa.oauth_private_key
```
mikeal: We shouldn't alter these variable names. They map to actual OAuth standard names so we should keep them.

We should stick `/* eslint-disable camelcase */` and `/* eslint-enable camelcase  */` around the code to just disable this for this code block.

ahmadnassri: on it.

ahmadnassri:
@mikeal updated.

I opted for line-level lint disable, rather than wrapping the whole block, as other variables are declared within that should follow the `camelCase` restriction _(future proofing)_

lib/oauth.js
ahmadnassri: this has been done ... not sure why github still shows this as unresolved.

ahmadnassri:
@FredKSchott done.

FredKSchott:
Looks like there are conflicts too, can you resolve?

ahmadnassri:
ah, didn't see on mobile (where I made that last commit) will rebase when on a desktop.

ahmadnassri:
@FredKSchott managed to do it on mobile 📱 🎉

FredKSchott:
Nice! 

Big lint PRs have a tendency to stagnate quickly, will merge asap (the issue discussing this has been open for a while now without any opposition so this shouldn't be catching anyone by surprise)

FredKSchott:
Thanks @ahmadnassri!

PR #2580, opened 2017-03-05
Use standard style

mikeal:
Phew!

`standard --fix` didn't handle a lot of our style miss-matches so this was a big manual effort to get going.

All tests work except the karma tests because it doesn't suppose `const` declarations.


mikeal:
Removing PR, there's another PR that is better.

Issue #2582, opened 2017-03-06
Allow to set main content-type header with form/formData


kjona:
When sending a form, you can control the boundary, similarly to the advanced use case of forms from the `README.md`:

```js
var r = request({uri: 'http://service.com/upload', method: 'post'}, function optionalCallback(err, httpResponse, body) {...})
var form = r.form();
form._boundary = 'MY-BOUNDARY';
form.append('my_field', 'my_value');
...
```

The content type header of the request will be set to `'multipart/form-data; boundary=MY-BOUNDARY'`. 

However, I need to control the content type, e.g. `'multipart/MY-TYPE; boundary=MY-BOUNDARY'`.

This is not possible, although the `form-data` package allows for this. Reason is that when the `getHeaders()` method of the form object is invoked with an empty/undefined `userHeaders` argument.

(`form-data/lib/form_data, ll. 281ff.`)
```js
FormData.prototype.getHeaders = function(userHeaders) {
  var header;
  var formHeaders = {
    'content-type': 'multipart/form-data; boundary=' + this.getBoundary()
  };

  for (header in userHeaders) {
    if (userHeaders.hasOwnProperty(header)) {
      formHeaders[header.toLowerCase()] = userHeaders[header];
    }
  }

  return formHeaders;
};
```

(`request.js, ll. 571ff`)
```js
if (self._form && !self.hasHeader('content-length')) {
  // Before ending the request, we had to compute the length of the whole form, asyncly
  self.setHeader(self._form.getHeaders(), true)
  self._form.getLength(function (err, length) {
    if (!err && !isNaN(length)) {
      self.setHeader('content-length', length)
    }
    end()
  })
} else {
  end()
}
```

A simple solution would be to pass `self.headers` as `usedMembers` argument. I will submit a pull request for this.

PR #2583, opened 2017-03-06
Allow to set main content-type header with form/formData

kjona:
Request headers are passed to the `form-data` `getHeaders()` method, so it is possible to override the default header for `content-type`, generated by `form-data`.
For detailed description, see issue [Allow to set main content-type header with form/formData](https://github.com/request/request/issues/2582)

### DIFF
--- a/tests/test-headers.js
+++ b/tests/test-headers.js
@@ -203,6 +203,23 @@ tape('catch invalid characters error - POST', function(t) {
   })
 })
 
+tape('allow custom content type on form data - POST', function(t) {
+  request({
+    url: s.url + '/headers.json',
+    headers: {
+      'content-type': 'custom'
+    },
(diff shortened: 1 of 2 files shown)


tests/test-headers.js
```diff
@@ -203,6 +203,23 @@ tape('catch invalid characters error - POST', function(t) {
   })
 })
 
+tape('allow custom content type on form data - POST', function(t) {
+  request({
```
Huachao: also specify the method `POST`

kjona: done

BlueBird67:
Hello, the solution from @Huachao is working great, because else it is not working at all when you use API Web Services that verify Content-headers...

When this will be integrated in a new version?

PR #2592, opened 2017-03-12
New option endRequest which can turn off automatic request closing

ovx:
## PR Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -749,6 +749,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 - `json` - sets `body` to JSON representation of value and adds `Content-type: application/json` header. Additionally, parses the response body as JSON.
 - `jsonReviver` - a [reviver function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) that will be passed to `JSON.parse()` when parsing a JSON response body.
 - `jsonReplacer` - a [replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) that will be passed to `JSON.stringify()` when stringifying a JSON request body.
+- `endRequest` - whether to close the request automatically (default: true).
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -749,7 +749,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 - `json` - sets `body` to JSON representation of value and adds `Content-type: application/json` header. Additionally, parses the response body as JSON.
 - `jsonReviver` - a [reviver function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) that will be passed to `JSON.parse()` when parsing a JSON response body.
 - `jsonReplacer` - a [replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) that will be passed to `JSON.stringify()` when stringifying a JSON request body.
-- `endRequest` - whether to close the request automatically (default: true).
+- `keepOpen` - whether to keep the request open for future writes (default: false).
 
 ---
 
--- a/request.js
+++ b/request.js
@@ -523,8 +523,8 @@ Request.prototype.init = function (options) {
     }
 
     var closeRequest = function() {
-      if (self.endRequest !== false) {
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -749,6 +749,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 - `json` - sets `body` to JSON representation of value and adds `Content-type: application/json` header. Additionally, parses the response body as JSON.
 - `jsonReviver` - a [reviver function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) that will be passed to `JSON.parse()` when parsing a JSON response body.
 - `jsonReplacer` - a [replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) that will be passed to `JSON.stringify()` when stringifying a JSON request body.
+- `endRequest` - whether to close the request automatically (default: true).
```
FredKSchott: `endRequest` isn't very clear. What about `keepOpen`? `closeAfterResponse`?

request.js
```diff
@@ -577,7 +577,8 @@ Request.prototype.init = function (options) {
         }
         end()
       })
-    } else {
+    } else if (self.endRequest !== false) {
```
FredKSchott: what about all the other places where end() & self.end() are called? Should those be blocked from automatically closing as well?

tests/test-endRequest.js
```diff
@@ -0,0 +1,60 @@
...
+   
+  stream = request.post({
+    url: server.url
+  }, function(err, res, body) {
+    stream.on('error', function(err) {
+      t.equal(err.message, 'write after end')
+      t.end()
+    })
+
+    stream.write('test data')
```
FredKSchott: this test will still pass if no error is emitted from this. Change test so that it fails if error is never emitted.

tests/test-endRequest.js
```diff
@@ -0,0 +1,60 @@
...
+    t.equal(res.statusCode, 200)
+    t.equal(body, testData + testData)
+    t.end()
+  })
+
+  setTimeout(function() {
+    stream.write(testData)
+  }, 10)
+
+  setTimeout(function() {
```
FredKSchott: don't rely on timing here. Try to write after ending and assert that no error was thrown and that data was sent in the request.

ovx:
> Doesn't end call the callback? You tests call end inside the callback, so I'm confused here.

the callback should be executed once the response stream ends; in the tests, there's `t.end()` which is supposed to end the test, not the request

>Why use this instead of piping a stream into request? When the stream interface is used, end() is called manually by the read stream (automatically called when using pipe()).

Piping a stream is fine, but it requires a stream which in some cases (like writing manually data using .write()) would require a proxy stream. This solution simply provides a more convenient way of working with such streams.

> Using a callback and streaming data with write() will cause problems, you should use one or the other. (You should listen to the response and error events if you want to use the stream interface.)

What kind of problems? This is exactly what I'm trying to achieve here because relying on additional proxy stream is more complex in many cases and the solution proposed here would make this simpler and cleaner.

> I think there's a chance you've misunderstood how to stream data into the request. If you don't include a callback, you should see the behavior you're looking for.

The request stream is closed with setImmediate even when you don't provide a callback. This solution proposed here prevents this from happening and allows you to work with the request stream any time later.

Consider the following scenarion:

```js
let stream = request.post('http://...');

setImmediate(() => {
  // this throws because the request stream is already closed
  stream.write('data');
});
```

You can't write to the request stream because it is already closed. You can only write to that stream within the same tick. The solution proposed here allows you to write to the request stream many ticks later.

The same applies for pipes:

```js
let stream = request.post('http://...');

setImmediate(() => {
  // this throws because the request stream is already closed
  fs.createReadStream('...').pipe(stream);
});
```

README.md
ovx: `keepOpen` is much better name indeed

request.js
ovx: Moved to the local end() function where it is indeed better.

tests/test-endRequest.js
ovx: Can't write after ending because the sole purpose of this test is to test manual ending after a few ticks. SetTimeouts replaced with setImmediate.

tests/test-endRequest.js
ovx: This test will timeout if no error is emitted; but changed

PR #2594, opened 2017-03-14
chore(dependencies): har-validator to 5.x [removes babel dep]

ahmadnassri:
## PR Checklist:

## PR Description
a simple `npm install request` resulted in a `babel` install due to usage as a `devDependency` in `har-validator`, so `har-validator` is now refactored to remove that dependency entirely, making `request` lighter to install _(if only slightly)_ :)


FredKSchott:
Nice! Thanks for cleaning that up for everyone

PR #2604, opened 2017-03-21
Improvemts in digest auth implementation

albertsuchvi:
## PR Checklist:
- [x ] I have run `npm test` locally and all tests are passing.
- [ x] I have added/updated tests for any new behavior.
       
- [x ] If this is a significant change, an issue has already been created where the problem / solution was discussed: https://github.com/request/request/issues/2546
       

## PR Description
As described in issue 2546, this PR tries to address a couple of limitations of current implementation of digest authentication in request:

1.- There is no way to ensure that digest authentication is the only authentication mechanism that can be used. It is possible for an attacker to do a MitM and 'downgrade' teh authentication to basic and thus capturing user credentials.
To address this problem I have added an additional (optional) field to the auth parametrer that can disable specifc authentication mechanisms when setting the immediate parameter to false. Thus, the client app can disable basic (and bearer) authentication. When the server responds with the authenticate header, if the authentication requested is disabled by the client, the request fails with an error.

2.- Reusing cnonces and handling nc
When issuing a request with digest auth, the previous cnonce was not used. The consequence is that every request with digest auth had to perform two http requests (one to get the cnonce, and another to send teh digest authentication data using the new cnonce). 
I have improved this keeping the last cnonce and trying to reuse it when a request (to the same server) with immediate set to false is performed. This can reduce the number of http requests actually sent in teh case that several accesses to the same server are perfomed consecutively. This improvement also required ensuring that a different nc is sent with each request containg digest auth.


albertsuchvi:
Error in lint, will retry when everything is fixed. It seems there was a problem in teh config of lint on my side and I ddi not see the errors b4 PR

PR #2613, opened 2017-03-27
Documentation: Added Simple JSON Example with oAuth

sgelbart:
I think this is a common enough example to warrant it own example. This assumes you are doing oAuth with your known token/keys (connecting to your own app). Not sure the best way to word the difference between that and your other example...

## PR Checklist:
       
### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ request('http://www.google.com', function (error, response, body) {
 
 ## Table of contents
 
+- [Simple JSON with oAuth](#simple-json-with-oauth)
 - [Streaming](#streaming)
 - [Forms](#forms)
 - [HTTP Authentication](#http-authentication)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,7 @@ request('http://www.google.com', function (error, response, body) {
 
 ## Table of contents
 
+- [Simple JSON with oAuth](#simple-json-with-oauth)
```
Huachao: I think `OAuth` is more general than `oAuth`

README.md
```diff
@@ -46,6 +47,33 @@ lots of [usage examples](#examples) and several
...
+        'X-Auth-Token': 'your-auth-token-id', //update this with your auth token
+        'Content-Type': 'application/json',
+        'Accept': 'application/json',
+      },
+      json: {
+        "some-data": "some-value" //add your data here
+      }
+    }, 
+    function (error, response, body) {
+      if (!error && response.statusCode == 200) {
```
Huachao: ===

README.md
```diff
@@ -46,6 +47,33 @@ lots of [usage examples](#examples) and several
...
+        console.log(body); //handle success
+      }
+      else {
+        console.log("Error "+response.statusCode);  //handle error
```
Huachao: add space before and after '+'

sgelbart:
Don't know if this is still relevant/accurate but just noticed the comments so updated in case it is

PR #2636, opened 2017-04-13
fix logic

minionymous:
old code: !self.pool && self.pool !== false
+ self is always defined, so self.pool has 2 values: undefined or defined
+ self.pool === undefined: undefined type is different with boolean so statement 2 will TRUE
+ self.pool === false: self.pool !== false will make the test fail, again
+ self.pool === something_else: !self.pool will make test fail
====
I think at this line it should compare self.pool with undefined

## PR Checklist:
       
### DIFF
--- a/request.js
+++ b/request.js
@@ -169,7 +169,7 @@ Request.prototype.init = function (options) {
   self._qs.init(options)
 
   debug(options)
-  if (!self.pool && self.pool !== false) {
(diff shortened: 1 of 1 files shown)


FredKSchott:
Sorry, `{pool: false}` is a valid option that we need to handle specially here. See the README & `test-pool.js` for an example of this option in action.

PR #2638, opened 2017-04-13
add socks5 support

minionymous:
## PR Checklist:
       

FredKSchott:
Hi @minionymous, thanks for your contribution! Unfortunately we cannot add extra dependencies to the request project to support this one option. If you'd still like to see this support, I'd recommend writing and publishing your own request wrapper for anyone who wants socks5 support. 

If you do, please come back and link to it here so that anyone searching can find it easily.

PR #2541, opened 2017-02-14
Add convenience method for HTTP OPTIONS

jamesseanwright:
Hey guys,

I'm currently working on a feature that involves a HTTP OPTIONS request. I really appreciate the convenience methods as they are easy to stub with my library of choice (Sinon).

I noticed that there is not an equivalent for HTTP OPTIONS, so I've added one. There may be a reason why this hasn't been exposed already, but implementing it didn't impact any of the existing tests, plus searching the existing PRs and issues didn't suggest anything.

My PR also has new version number (i.e. new minor revision), but this is merely a suggestion; I'm sure you'll have your own processes surrounding this.

Let me know if you have any questions or thoughts.

Thanks,
James

PR #2641, opened 2017-04-17
Update README to simplify & update convenience methods

FredKSchott:
## PR Checklist:
       
- [X] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A
       

## PR Description

I realized that there is a README section on the request convenience methods after merging #2541, so I made a quick update to that section to simplify it (no need for it to be so long) and update it (add the newest `request.options()` method).

/cc @jamesseanwright

### DIFF
--- a/README.md
+++ b/README.md
@@ -891,55 +891,19 @@ var specialRequest = baseRequest.defaults({
 })
 ```
 
-### request.put
+### request.METHOD()
 
-Same as `request()`, but defaults to `method: "PUT"`.
+These HTTP method convenience functions act just like `request()` but with a default method already set for you:
 
-```js
-request.put(url)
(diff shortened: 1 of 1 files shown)


jamesseanwright:
Oops, sorry I missed this! Thanks for the update.

tsuzfr:
Thanks for merging the update. However, v2.81.1 does not seem to be in the npm directory. The latest one is v2.81.0. Do you think we can update this?

FredKSchott:
That's intentional to differentiate between the version on npm & the version on github/master

nylen:
The only times the .0 vs .1 thing has ever come up, it has been more confusing than helpful.

PR #2644, opened 2017-04-18
Update README.md

deed02392:
## PR Checklist:
- [N/A] I have run `npm test` locally and all tests are passing.
- [N/A] I have added/updated tests for any new behavior.
       
- [N/A] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A, or add link to issue here]
       

## PR Description
Added a paragraph of text to the TLS section, on the use of the `ca` option, explicitly detailing how to get SSL certificates that consist of more than one issuer to successfully verify. It was not made clear that `ca` can be an array, and that in fact this is necessary (AFAICT) when you want to accept a certificate that has been signed by multiple issuers, common in internal environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,8 +677,8 @@ request.get({
 
 The `ca` value can be an array of certificates, in the event you have a private or internal corporate public-key infrastructure hierarchy. For example, if you want to connect to https://api.some-server.com which presents a key chain consisting of:
 1. its own public key, which is signed by:
-2. an intermediate **some-server Corp Issuing Server 01**, that is in turn signed by: 
-3. a root CA **some-server Corp Root CA**;
+2. an intermediate "Corp Issuing Server", that is in turn signed by: 
+3. a root CA "Corp Root CA";
 
 you can configure your request as follows:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -675,6 +675,22 @@ request.get({
...
+The `ca` value can be an array of certificates, in the event you have a private or internal corporate public-key infrastructure hierarchy. For example, if you want to connect to https://api.some-server.com which presents a key chain consisting of:
+1. its own public key, which is signed by:
+2. an intermediate **some-server Corp Issuing Server 01**, that is in turn signed by: 
+3. a root CA **some-server Corp Root CA**;
+
+you can configure your request as follows:
+
+```js
+request.get({
+    url: 'https://api.some-server.com/',
```
FredKSchott: Some of this is confusingly worded, I'd recommend simplifying to:
```
1. its own public key, which is signed by:
2. an intermediate *Corp Issuing Server*, that is in turn signed by: 
3. a  *Corp Root CA*.
```

Note the switch away from bold, which isn't needed to draw attention here. Switch to either italics or double quotes.

deed02392:
I wanted to make it simple for viewers to look between the listed certificates and the filenames given in the array, because the order they appear in the array is so important.

I will compromise with you by removing bold, and formatting the array declaration over lines with inline comments.

deed02392:
Please merge and close.

PR #2647, opened 2017-04-24
Support asynchronous followRedirect filter (callback or Promise)

brianreavis:
## PR Checklist:

## PR Description

This can be needed for hostname to ip resolution, which can be used to disallow redirection to internal network addresses for protection against SSRF attacks. This PR also allows the `followRedirect` result to be a string, which overrides the redirect location. This can be used translate the target hostname to a ip (locking it).

```js
// using a callback
followRedirect: function(response, callback) {
    callback(null, result);
}

// using a promise
followRedirect: function(response) {
    return new Promise(...);
}
```

After writing this I ended up just using [`paranoid-request`](https://github.com/uber-common/paranoid-request) for my use case – so take this PR with a grain of salt. That said, implementing this sort of behavior with normal `request` would be non-trivial / not possible without this PR or something similar (unless I'm missing something obvious).

### DIFF
--- a/README.md
+++ b/README.md
@@ -777,7 +777,10 @@ The first argument can be either a `url` or an `options` object. The only requir
 
 ---
 
-- `followRedirect` - follow HTTP 3xx responses as redirects (default: `true`). This property can also be implemented as function which gets `response` object as a single argument and should return `true` if redirects should continue or `false` otherwise.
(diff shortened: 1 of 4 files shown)


RamprasadAkella:
Can you guys please merge this PR or add this functionality to the request module ?

PR #2664, opened 2017-05-20
Lazy loading tough-cookie

chriswoodle:
## PR Checklist:

## PR Description
I'm currently working on a device with low memory and while going through heapdumps, one of Request's deps, [tough-cookie](https://www.npmjs.com/package/tough-cookie), uses a lot of memory by [loading a large array](https://github.com/salesforce/tough-cookie/blob/master/lib/pubsuffix.js) of domain name suffices when using `require`. This PR lazy loads the `tough-cookies` module and components for handling cookies. Now modules using Request that do not use cookies will not load `tough-cookies` into memory.

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -1,10 +1,5 @@
 'use strict'
 
-var tough = require('tough-cookie')
-
-var Cookie = tough.Cookie
-var CookieJar = tough.CookieJar
-
 exports.parse = function (str) {
   if (str && str.uri) {
     str = str.uri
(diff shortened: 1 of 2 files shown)


ZJONSSON:
calling require is not free, even though the module is loaded.  Example:
```js
require('tough-cookie');
console.time('require');
for (let i = 0; i < 500000; i++) {
  require('tough-cookie');
};
console.timeEnd('require');
//  prints require: 823.934ms
```

For lazy loading it might be better to define an unassigned variable and only require touch cookie if the variable is still unassigned, i.e.:

```js
let toughCookie;

exports.parse = function (str) {
  ...
  toughCookie = toughCookie || require('tough-cookie')
  return toughCookie.Cookie.parse(str, {loose: true})
}
```

Issue #2683, opened 2017-06-07
Failure to set proxy during redirects


atti187:
### Summary
I have a scenario where a request is being made to an internal company server (no proxy) which redirects to an external server (proxy needed). The consecutive request is also being done without a proxy.

### Expected Behavior
The redirect request should validate if a proxy is needed, based on the url

### Current Behavior
No additional proxy information is added to the request

### Possible Solution
https://github.com/request/request/compare/master...atti187:master
 
### Context
An example is how Microsoft TFS handles npm-feeds, see https://www.visualstudio.com/en-us/docs/package/npm/upstream-sources#caching (Internet requirements). Normally, TFS is installed on-prem (no proxy needed) but will redirect to npmjs.org (proxy needed).

### Your Environment
| npm | 4.6.1
| ---------------- | -------
| request          | 2.81.1
| node             |v6.10.0
| Windows | Windows 7 Enterprise

PR #2685, opened 2017-06-08
Failure to set proxy during redirects

atti187:
## PR Checklist:
- [] If this is a significant change, an issue has already been created where the problem / solution was discussed: [#2683]

## PR Description
This change makes sure that the proxy setting is re-evaluated according to the redirect url


BYK:
@mikeal - Anything blocking the merge of this PR? It looks well-written, complete with a test case and merging that would allow us to update the requests version in Yarn and fix a bug instead of having implementing a workaround ourselves.

Issue #2708, opened 2017-07-04
S3 signature version 4: 'x-amz-content-sha256' header missing


paulrutter:
### Summary
When using 'sign_version': 4 when using S3, the following error is returned by AWS:
HTTP 400 "Missing required header for this request: x-amz-content-sha256".

### Simplest Example to Reproduce
```js
request({
  url: 'https://<bucketname>.s3.eu-central-1.amazonaws.com',
  aws: {
    bucket: 'bucketname',
    key: 's3 key',
    secret: 's3 secret',
    sign_version: 4
  },
  qs: {
    prefix: 'filename'
  }
});
```

### Expected Behavior
HTTP 200 is returned, containing an XML response.

### Current Behavior
HTTP 400 Bad Request containing the following error:
'Missing required header for this request: x-amz-content-sha256'

### Possible Solution
In request.js -> Request.prototype.aws method, add the following after line 1395:
```js
self.setHeader('x-amz-content-sha256', signRes.headers['X-Amz-Content-Sha256'])
```
We would also need to update the 'aws4' module to the latest version.
I've tested with 1.6.0.

See http://docs.aws.amazon.com/AmazonS3/latest/API/bucket-policy-s3-sigv4-conditions.html and https://github.com/aws/aws-sdk-go/issues/743.

### Context
Connect to S3 by using signature version 4.

### Your Environment

| software         | version
| ---------------- | -------
| request          | 2.81.0
| node             | 6.9.1
| Operating System | Windows

PR #2709, opened 2017-07-04
Add proper S3 signature version 4 support

paulrutter:
Add proper S3 signature version 4 support.
See https://github.com/request/request/issues/2708 for more info.


dominikkadera:
Why isn't this still merged? I'm working with Amazon S3 right now and I need this.

paulrutter:
@dominikkadera i finally proceeded by using https://aws.amazon.com/sdk-for-node-js/. The is the official library from AWS for handling S3 (and other). It offers far more features (streaming files for example, which can be handy for large files) and supports all regions by default.
I would advice to use that module instead, when you want proper S3 support.

Issue #1502, opened 2015-03-23
Cookies lost during multiple redirects


nijikokun:
1. Put a cookie on a 302
2. Do another 302
3. View headers, cookie from the first 302 does not exist.

lalitkapoor:
Hey Niji, can you give me more details. What were the domains and protocols of the redirects? Thanks.

lalitkapoor:
Hey @Nijikokun can you give me more details. What were the domains and protocols of the redirects? Thanks.

Nijikokun:
Can't give the domains as they are internal, should work with any, https.

Cookie type was: `http-only; secure`

lalitkapoor:
thanks! Will look into it. If it's not too much trouble could you create a failing test?

anklos:
I tried with node `https` module, it does not preserve cookie from first redirect either.

lalitkapoor:
@anklos you could set request to not follow redirects and do the redirect manually, in which case you could get the first response.

anklos:
@lalitkapoor thanks. I set `followRedirect` to false in the option. It seems that I could reach the first response, as the statusCode is 302, however, the header is the same as after redirects response's header, it does not contain `set-cookie` content from the first redirect.

![image](https://cloud.githubusercontent.com/assets/530145/6777142/bab3d244-d19a-11e4-9c15-4cbc9af23316.png)

Nijikokun:
Cookies are not passed with redirects

lalitkapoor:
@anklos can you tell me if the domains are different?

lalitkapoor:
I don't need the exact domains, just need to know if they're the same...if one is a subdomain, etc.

anklos:
they are on the same subdomain.

lalitkapoor:
@anklos are you able to modify the cookie on the servers? or do you not have access to do that. I'd like to try something: setting the domain value on the cookie to the hostname (w/subdomain).

@Nijikokun do your cookies have the domain value set on them?

anklos:
@lalitkapoor no I dont. but I can try to set up a https server myself to test the cookie behavior.

anklos:
although still not make sense to me why browser could get the cookies but not my node.js program?

anklos:
Just find that I can get `set-cookie` by setting redirect option false. Originally I did not get it because I send request to the first redirect url, not the destination url. 

Also tried other language http libraries to do the same test, the results are the same as nodejs

lalitkapoor:
[...]

PR #2710, opened 2017-07-05
Create jar if it's undefined

kyleholzinger:
## PR Checklist:
- [x] I have run `npm test` locally and all tests are passing. (Had failing specs before, no new ones failed)
       
## PR Description

This is in response to #1502, I'm not super familiar with the in's and out's of `request`, but hopefully this helps!

### DIFF
--- a/request.js
+++ b/request.js
@@ -295,7 +295,7 @@ Request.prototype.init = function (options) {
     self.setHost = true
   }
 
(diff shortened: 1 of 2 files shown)


mikeal:
Does this enable a jar by default? We haven't done this in the past, so it's a change in defaults, and could be considered a major breaking change.

PR #2725, opened 2017-07-27
Better example for "Custom HTTP Headers"

eklem:
## PR Checklist:
       
## PR Description
Only changed a README.md example for [Custom HTTP Headers](https://github.com/request/request#custom-http-headers), moving the url out of config. The reason for this is that it's a better example to build upon, since options are mostly static, and URLs to crawl not so much.


eklem:
I guess this is not happening 😄

PR #2741, opened 2017-08-15
feat(aws): allow service and region to be specified in aws options.

chasingmaxwell:
## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A (small change)

## PR Description
This PR picks up where #2188 and #2339 left off to allow aws signing for API Gateway requests which require service and region as part of the signature. aws4 supports adding both `service` and `region` so this is a simple change.

Many thanks to @NewGyu who originally wrote this code which I've simply applied to the latest master!


chasingmaxwell:
apologies. I did run tests locally, but there appears to be a problem. I'm looking into it.

chasingmaxwell:
There we go!

request.js
```diff
@@ -1364,6 +1364,8 @@ Request.prototype.aws = function (opts, now) {
       host: self.uri.host,
       path: self.uri.path,
       method: self.method,
+      service: opts.service,
```
elliotttf: Is it expected that custom domains would only be available for version 4 signing?

chasingmaxwell: I'm not positive what `service` and `region` might be used for other than API Gateway. API Gateway only supports version 4 though.

cmbrehm:
👍  for merging this, working around it currently.

johnz:
When can we get this merged? we need 'region' to sign AWS API Gateway url with custom DNS

crsepulv:
When this will be merged? I need to sign4 request to API Gateway, and the service param is needed to create the signature.

andrhamm:
Please merge

holmberd:
I see `service` for aws4 but no `region` https://github.com/request/request/blob/e81c9b1dadffe692ac6797fbf7e0bdba557ebfdc/request.js#L1369

ffoysal:
we need this feature. Please merge

PR #2751, opened 2017-08-24
Update of hawk and qs to latest version

Olivier-Moreau:
## PR Checklist:
       
## PR Description

I updated hawk from 3.1.3 to 6.0.2 (still a ~ for version managing)
I updated qs too  from 6.4.0 to 6.5.0 (kept the ~)

npm test ran without error. I didn't add a test since I only changed version.


louisbuchbinder:
@Olivier-Moreau @mikeal 

Hi sorry to be a pain, but this change broke semver minor version update. This is a breaking change as the new hoek uses `const` instead of `var`

```const Hoek = require('hoek');
^^^^^
SyntaxError: Use of const in strict mode.```

**Comment by lostip** [comment]
@louisbuchbinder the `package.json` has the engine properly set:
```
"engines": {
     "node": ">= 4"
   },
```
but npm has a very unfortunate default, where this is interpreted as a suggestion sort of, unless you use the `engine-strict` flag.
See: https://docs.npmjs.com/files/package.json#engines

PR #2771, opened 2017-09-18
Add promise support

utyfua:
This patch add promise support.
> request('https://www.google.com.ua/').then(response=>console.log(response))


mikeal:
Sorry, the return value from request is already a Stream so we can't support promises this way.

Checkout https://github.com/request/request-promise

PR #2802, opened 2017-10-28
Make 'uri is a required argument' error consistent

zaverden:
## PR Checklist:
       
## PR Description
There is an inconsistency in 'missing uri' errors.

1. Call `request` without `uri` and `baseUrl` (`request({ });`)
2. Call `request` without `uri` but with `baseUrl` (`request({ baseUrl: 'https://github.com' });`)

**Actual**:
The 1st call throws `options.uri is a required argument` error, the 2nd one throws `options.uri must be a string when using options.baseUrl` error

**Expected**:
Both call throw `options.uri is a required argument` error.

PR adds missing test and fixes the problem

### DIFF
--- a/request.js
+++ b/request.js
@@ -197,7 +197,7 @@ Request.prototype.init = function (options) {
 
   // If there's a baseUrl, then use it as the base URL (i.e. uri must be
   // specified as a relative path and is appended to baseUrl).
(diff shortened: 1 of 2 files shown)


zaverden:
Closed by #3142

Issue #2803, opened 2017-10-28
[Feature] Route-like format for URI


zaverden:
### Summary

I want to have an alternative to concatenation and string interpolation for dynamic URIs.

### Simplest Example to Reproduce

```js
request({
  baseUrl: 'https://github.com',
  uri: '/:user/:name/search?q=uriData',
  uriData: {
    user: 'request',
    name: 'request'
  })
```

### Expected Behavior

`request` should find parameters `:user` and `:name` in the URI and replace them with provided values from `uriData` object. Values should be encoded if needed.

### Current Behavior

`uriData` is ignored and URI will be used as provided

### Possible Solution

`koa` and `express` use `path-to-regexp` library to handle routes. The library also provides methods to [combine an URL from route and data](https://github.com/pillarjs/path-to-regexp#compile-reverse-path-to-regexp). Same syntax (colon-prefixed parameter) is used by `ui-router`. So front-end and back-end developers are familiar with this syntax.

I will provide a PR.

### Context

It is a common situation, when URI is not static but depends on some data. For example, I have a github repository model:
```js
const githubRepo = {
  user: 'request',
  name: 'request'
};
```
To get an URI for search request I need to use concatenation
```js
request('https://github.com/' + githubRepo.user + '/' + githubRepo.name + '/search?q=uriData');
```
or may to use string interpolation
```js
request(`https://github.com/${githubRepo.user}/${githubRepo.name}/search?q=uriData`);
```
which is better, but still is hard to read.
Route syntax is much clear:
```js
request('https://github.com/:user/:name/search?q=uriData', { uriData: githubRepo });

```
URL encoding is also an issue when developer manually combines an URI.

zaverden:
I still think this API will be helpful. Will try to provide another solution during NY holidays

reconbot:
Request is now in maintenance mode and wont be merging any new features. Please see #3142 for more information.

PR #2804, opened 2017-10-28
Add support route-like URI populated from `uriData`

zaverden:
## PR Checklist:
       
## PR Description

Solution for #2803

### DIFF
--- a/README.md
+++ b/README.md
@@ -682,7 +682,7 @@ request.get({
 
 ## Support for HAR 1.2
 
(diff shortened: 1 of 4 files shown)


mikeal:
Anything that expands use of regex needs a decent security review. Maybe @evilpacket has thoughts.

evilpacket:
While I haven't looked deep into `path-to-regexp` the situation we want to avoid is that module creating a catastrophic regex that when user input applied would create a dos situation. I don't know if that library protects against that, it sort of hints that, but it's not clear what the goals are for that lib regarding security of untrusted user input.

PR #2827, opened 2017-12-07
Describe casing for method option

sharils:
## PR Checklist:
- [n/a] I have added/updated tests for any new behavior.
       
- [n/a] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A, or add link to issue here]
       

## PR Description
The code base is treating HTTP method casing inconsistently. For example, [request.js:1389](https://github.com/request/request/blob/master/request.js#L1389) converts the method to all uppercase, while [index.js:L49](https://github.com/request/request/blob/master/index.js#L49) is only checking against all upper case. In this specific case, if `` head `` is the method, it could still by pass all the checks. Thus IMO, the easiest way to solve this is to document the casing. Personally, I'd prefer to make HTTP method case-insensitive as a feature, however, due to lack of knowledge about the code base, I'm not able to create an PR for that.

---
This change is [Reviewable](https://reviewable.io/reviews/request/request/2827)


sharils:
Hi @mikeal - Would you mind taking a look at this PR? I'd like to get this doc change merged. The reason I'm asking your help is because I don't know who to ask help from and I saw that the latest PR was merged by you. Thanks for your time.

Issue #2606, opened 2017-03-23
Add an option to whether to keep authorization header on redirection


Huachao:
In some cases authorization header is not wanted on redirection like AWS, while some cases use may want to preserve this, so can we have this flexibility and make this as an option?

Issue #2278, opened 2016-07-10
Removal of Authorization header is surprising


pangel:
I had requests getting 401'd because of the following:

https://github.com/request/request/blob/290111d4e94ac9f6c8d3abecbc8920c96d1b32be/lib/redirect.js#L137

I suggest adding an option (e.g. `keepAuthorization: bool`), or some way of keeping all header across redirects.

PR #2876, opened 2018-02-19
Adding redirectLocationTrusted option to retain authorization header on redirect domain change

gareth-robinson:
## PR Checklist:
  - new tests added into tests/test-redirect-auth.js
- [Y] If this is a significant change, an issue has already been created where the problem / solution was discussed:
  - Not a significant change, but has been mentioned a few times in the issues:
    - https://github.com/request/request/issues/2606
    - https://github.com/request/request/issues/2278
  
## PR Description
If a redirect location has a different domain than the original request (excluding port and protocol changes) then any authorization header is removed from the request. This is currently a deliberate choice in the code, to do the same as curl (for security reasons), as per this comment: https://github.com/request/request/blob/master/lib/redirect.js#L134

Curl does have an option --location-trusted for overriding that behaviour though: (https://curl.haxx.se/docs/manpage.html#--location-trusted) so this pull request is to add an equivalent option `redirectLocationTrusted` to request. If not supplied this option defaults to false, keeping the original behaviour.

In this change I've kept `redirectLocationTrusted` as the simple boolean that curl has. It _could_ be overloaded for finer grained options (e.g. a domain matching regex) if it was felt that would be of more value.


gareth-robinson:
Bleurgh, it worked locally with node 6.12 :( but travis CI is failing. Will investigate.

gareth-robinson:
Raised https://github.com/request/request/issues/2877 to discuss fixing the Travis CI issues

Issue #2067, opened 2016-02-11
toString failed - buffer size issue for big files (maxBufferSize)


theraaz:
I'm requesting a big file with 150K records but it throws 'toString failed' error. https://github.com/nodejs/node/issues/3175 says it's because of maxBufferSize. Request works well for 200 records but its an external api and the requirement is to get all records at once.[no pagination there :( ]
Is there any way to set buffersize for this request?

EDIT:

    request("http://www.site-containing-big-data/api",
			function (error, response, body) {
                console.log('got something to show');
				if(!error && response.statusCode == 200) {
					resolve(body);
				}else if(error){
					reject(error);
				}
			});

but nothing shows in console other than `toString failed` message

aprilandjan:
Maybe you can set the response encoding to binary data, otherwise request will try to transform the response data to string and this might fail.

Issue #2826, opened 2017-12-05
toString() Failed (Buffer)


iamcharles777:
buffer.js:513
    throw new Error('"toString()" failed');
    ^

Error: "toString()" failed
    at Buffer.toString (buffer.js:513:11)
    at Request.<anonymous> (/root/dan-phillips-web/node_modules/request/request.js:1137:39)
    at emitOne (events.js:96:13)
    at Request.emit (events.js:188:7)
    at IncomingMessage.<anonymous> (/root/dan-phillips-web/node_modules/request/request.js:1085:12)
    at IncomingMessage.g (events.js:291:16)
    at emitNone (events.js:91:20)
    at IncomingMessage.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:974:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)

lassmann:
I am having the same issue here. I am trying to make a post where one input has another request.
    const rp = require('request-promise');
    const options = {
    'url': 'http://urltomakepost.com',
    'headers': {
    'Content-Type': 'application/json',
    }
}
    options['formData'] = {
        media: {
            value: rp.get(fileUrl),
            options: {
                filename: 'preview.mp4'
            }
        }
    };

    return rp.post(options)
        .then(response => {
            console.log(resonse);
        })
        .catch(error => console.log(error));

watching on port 3000
buffer.js:556
  if (encoding === undefined) return buf.utf8Slice(start, end);
                                         ^

Error: "toString()" failed
    at stringSlice (buffer.js:556:42)
    at Buffer.toString (buffer.js:633:10)
    at Request.<anonymous> (/home/ubuntu/filemanager/aws-file-manager/node_modules/request/request.js:1137:39)
    at emitOne (events.js:116:13)
    at Request.emit (events.js:211:7)
    at Request.source.emit (/home/ubuntu/filemanager/aws-file-manager/node_modules/delayed-stream/lib/delayed_stream.js:30:21)
    at IncomingMessage.<anonymous> (/home/ubuntu/filemanager/aws-file-manager/node_modules/request/request.js:1085:12)
    at Object.onceWrapper (events.js:313:30)
    at emitNone (events.js:111:20)
    at IncomingMessage.emit (events.js:208:7)
    at endReadableNT (_stream_readable.js:1056:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
[...]

PR #2903, opened 2018-04-03
Handled '"toString()" failed' uncaught error.

aniltallam:
Fix for the issue - https://github.com/request/request/issues/2067

## PR Checklist:
      test case was not added because, huge files (>256mb) should be used for testing. attaching this file to the project will drastically increases the repo size.

## PR Description
For response size >256mb, buffer.toString throws an error (https://github.com/nodejs/node/issues/3175). Since this was thrown in end event handler, there is no way for the developers to catch these exceptions, so it is resulting in uncaught Errors/Exceptions. 

So we have to gracefully handle those scenarios by catching those errors and then emitting error event.

### DIFF
--- a/request.js
+++ b/request.js
@@ -1137,6 +1137,7 @@ Request.prototype.readResponseBody = function (response) {
         try {
           response.body = response.body.toString(self.encoding)
         } catch (e) {
+          e.code = 'NODE_RESPONSE_SIZE_LIMIT_REACHED'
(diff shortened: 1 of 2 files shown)


SharonR1009:
You may can try to upgrade to Node 10.0.0 to catch the error
https://github.com/request/request/issues/2826#issuecomment-387318734

Issue #2894, opened 2018-03-13
header field "host" includes port 80


zhaotian2470:
As the following code:
[https://github.com/request/request/blob/bbb3a0bbbe173342de8462a2171765a39681b248/request.js#L294](https://github.com/request/request/blob/bbb3a0bbbe173342de8462a2171765a39681b248/request.js#L294) 
If URI is "http://myhost:80/path", header field "host" will be set as "myhost:80".
Is it correct?

Note: before commit ff6d6c6e7a3b2c36618b5d1db662e10c929696e3, "host" header field will be set as "myhost"

paambaati:
Recently, most clients (including curl) and browsers have started to drop the `:80` & `:443` port suffixes in the Host header, and include it only when it is neither of the two. Dropping the port prefix for known services is still RFC-compliant; see http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

Examples of other projects stripping the port suffix —

1. Ruby - https://github.com/ruby/ruby/blob/d46336e71731aa64d71d4573b2741b7de43ec340/lib/net/http/generic_request.rb#L18
2. waterlink/rack-reverse-proxy#3
3. twilio/twilio-node#151
4. https://github.com/bokeh/bokeh/pull/3513
5. https://github.com/websocket-client/websocket-client/pull/248
6. https://stackoverflow.com/a/19169664

Related `request` issue — https://github.com/request/request/issues/515

paambaati:
@zhaotian2470 This can be closed now, as v2.88.1 fixes this.

PR #2904, opened 2018-04-04
#515, #2894 Strip port suffix from Host header if the protocol is known.

### LINKED ISSUE #515: Port 80 is included in the 'Host' request field (instead of defaulted) when it isn't with other popular HTTP clients
Hi,

So we encountered an interesting problem today, less so a bug with request but more of a incongruent behavior from request when compared against other browsers and client's we've tested. 

When working with cloudfoundry.com we found that their proxy did not consider a request to port 80 the same if the port was explicitly specified in the 'Host' header field.

So a request to http://foo.cloudfoundry.com:80/ and http://foo.cloudfoundry.com/ behaved differently. We narrowed it down to the difference in the Host field. Request included the port when explicitly specifying it, but not when it wasn't. 
- http://foo.cloudfoundry.com:80/ has a Host header of "Host: foo.cloudfoundry.com:80"
- http://foo.cloudfoundry.com/ has a Host header of "Host: foo.cloudfoundry.com"

No big deal right? Well so everyone else strips port 80 because it is the default :(
- Curl strips port 80 from the Host header
- wget strips port 80 from the Host header
- Safari strips port 80 from the Host header
- Firefox strips port 80 from the Host header
- Chrome strips port 80 from the Host header

You can test the behavior using "http://helloworld-ng.cloudify.com:80/" and "http://helloworld-ng.cloudify.com/" - notice that both URLs work the same in all of the above listed browsers

In curl we can see the failure case

``` shell
pair@foo:~pair/$ curl http://helloworld-ng.cloudfoundry.com:80/ -H "Host: helloworld-ng.cloudfoundry.com:80"
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
pair@foo:~pair/$ curl http://helloworld-ng.cloudfoundry.com:80/ -H "Host: helloworld-ng.cloudfoundry.com"
Hello from Cloud Foundry
```

Now let's try it with request

``` javascript

request=require('request');

request.get("http://helloworld-ng.cloudfoundry.com:80/",function(err,req,body){
  console.log("this will fail with port 80",body);
});

request.get("http://helloworld-ng.cloudfoundry.com/",function(err,req,body){
  console.log("this will not fail without port 80",body);
});

```

(We've asked cloudfoundry to fix this issue, but to be clear this is incongruent with every other client and browser we've tested) 

So then the question becomes - should request do the same? Sadly it seems like this shouldn't be an issue but it is an unexpected behavior and no one else does it - hence why I think request should strip port 80 from the Header field if it is specified. 

On that note node-http-proxy suffers form the same problem and a nice bit of logic that it uses some places to address it:

``` javascript
  var host=host+(port-80===0?"":":"+port)
```

paambaati:
@mikeal This PR should fix #515 and #2894.

## PR Checklist:
       
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: #515, #2894 

## PR Description
It drops the port suffix in the `Host` header for standard protocol + port combinations (HTTP + `:80` and HTTPS + `:443`). It partially reverts ff6d6c6e7a3b2c36618b5d1db662e10c929696e3, while still working for IPv6 addresses.

CC: @JamesMGreene, @simov


paambaati:
@mikeal With 237231e, I've included tests for the changes 🎉

paambaati:
Tests seem to be failing because of the `# IPv6 Host header` test - this is probably because Travis CI does not allow IPv6 addresses — see https://github.com/travis-ci/travis-ci/issues/4964.

paambaati:
@mikeal, @simov **Bump!**

paambaati:
@mikeal @simov This is also very similar to the recent NPM 418 errors discussed [here](https://news.ycombinator.com/item?id=17175960) and here - https://github.com/npm/npm/issues/20791

paambaati:
@mikeal Would you mind reviewing/merging this?

mikeal:
This must be a regression, we used to take care of this by using the property from url.parse that doesn't include the header on known protocols. Any idea when this might have happened?

Can you disable the IPv6 test since it doesn't work in Travis?

paambaati:
@mikeal I just checked, and the last CI runs are all green.

> Any idea when this might have happened?

From what I can tell, this happened in #2571, here - https://github.com/request/request/pull/2571/files#diff-ccc0734f65dd7a299409ff07d35be095L293

mikeal:
Ah, thanks, too bad that regression snuck in :(

JamesMGreene:
Sorry, guys!

PR #2914, opened 2018-04-17
fix: add nsp as a test pre-condition for added security

hvolschenk:
## PR Checklist:

## PR Description

Added NSP to the `npm test` command list so vulnerable dependencies cannot get merged in.

> Prevention ... cure

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "isstream": "~0.1.2",
     "json-stringify-safe": "~5.0.1",
     "mime-types": "~2.1.17",
+    "nsp": "^3.2.1",
     "oauth-sign": "~0.8.2",
     "performance-now": "^2.1.0",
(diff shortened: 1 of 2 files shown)


hvolschenk:
Not sure how "significant" this change is viewed as, so I'll leave that checkbox until a bit later, or until a contributor checks it or tells me otherwise :)

hvolschenk:
I also see that `nsp` is not working in **Node4**: https://travis-ci.org/request/request/jobs/367527189. That's no good.

It also seems like the `http` error message for invalid headers is different on _Travis_ than it is on my machine (`Invalid character in header content ["test"]` on _Travis_, and `The header content contains invalid characters` on my machine). Will have to investigate this later. If the **Node4** thing is okay by you.

PR #2915, opened 2018-04-18
fix unixsocket when url contains colon

willkan:
## PR Checklist:
       
- [X] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A, or add link to issue here]
       

## PR Description

There is a bug when request unix url with search which contains `:`. This pr is expecting to fix this bug.

for example `http://unix:/absolute/path/to/unix.socket:/request/path?x=a:b`, will only send  `http://unix:/absolute/path/to/unix.socket:/request/path?x=a` to the unix socket server


willkan:
merge from the new master.

willkan:
@mikeal could u check this pr and merge it if it is ok?

Issue #2877, opened 2018-02-20
Tests consistently failing in Travis CI


gareth-robinson:
### Summary
Since the end of 2017 all Pull requests are all failing at the test-headers test; this appears to be because of two issues
1. The 'catch invalid characters error' test works on node v4 & v6, but fails on v8.
2. The 'IPv6 Host Header' test fails in all builds

### Simplest Example to Reproduce
These are test bugs rather than functional bugs - compare the CI builds for the recent pull requests and see they all fail in the same way.

#### 1. Headers characters tests fails on node v8
The 'catch invalid characters error' test works on node v4 & v6, but fails on v8.
This test appears to be asserting _node_ functionality; there isn't anything in the request code that actually throws the error message the test looks for. The error message that node throws has changed between versions:

v6 - https://github.com/nodejs/node/blob/v6.x/lib/_http_outgoing.js#L359
v8 - https://github.com/nodejs/node/blob/master/lib/_http_outgoing.js#L496

So what's the best solution here?

1. Remove the test, since it is checking node functionality rather than the library functionality?
Or if the test needs to exist to check that `request` properly rethrows the error
2. Change request to catch the error and throw it's own consistent error?
3. Change the test to check accept either error message?

I'd prefer 3. here I think.

#### 2. The 'IPv6 Host Header' test fails in all builds
There are a number of pages on this e.g.
https://github.com/travis-ci/travis-ci/issues/4964#issuecomment-159748141
https://github.com/travis-ci/travis-ci/issues/9207
https://github.com/njh/travis-ipv6-test

Basically it appears that IPv6 is not supported on travis-ci, and only worked previously as 'a by-product of where we've been hosting' (first link above). The second link above suggests it stopped working around the end of October, which would match when the pull requests started failing.

So what's the best solution here? I've only got one solution, so others are welcome...

1. Since the environments don't support IPv6, then don't run the IPv6 test. Pass some environment flag that allows that particular test to be skipped.

### Context
Blocking pull requests :)

### Your Environment
travis-ci

gareth-robinson:
Created pull request https://github.com/request/request/pull/2923 with alterations for this bugs

simov:
Fixed in: https://github.com/request/request/pull/2923

PR #2923, opened 2018-04-22
Alterations for failing CI tests

gareth-robinson:
## PR Checklist:
- [Y] If this is a significant change, an issue has already been created where the problem / solution was discussed: https://github.com/request/request/issues/2877

## PR Description
This is a suggestion of alterations to test-headers.js to unblock CI again, given there are a number of consistently failing tests there for some months now.
I've changed the invalid header characters test, so that it passes with _either_ the old or new message returned by node.
I've altered the IPv6 header test, so that if the underlying OS doesn't support IPv6 the test doesn't run.

### DIFF
--- a/tests/test-headers.js
+++ b/tests/test-headers.js
@@ -5,6 +5,15 @@ var request = require('../index')
 var util = require('util')
 var tape = require('tape')
 var url = require('url')
+var os = require('os')
+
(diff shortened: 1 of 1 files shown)


simov:
Thanks @gareth-robinson :+1:

PR #2943, opened 2018-05-16
Replace hawk dependency with a local implemenation

### LINKED ISSUE #2831: Remove hawk dependency
I can't take this any more.

Request has a dependency on [hawk](https://www.npmjs.com/package/hawk) even though no one uses that feature. It was added because @mikeal was using it a few years ago. People who use hawk are more likely to be using other hapi ecosystem modules like wreck and no hawk user needs it integrated like this. I wrote hawk and am its main user and I don't want it integrated in wreck, the module I use instead of request.

Every other week some request user opens an issue on hawk, hoek, boom, or the other dependencies with shitty attitude and a sense of entitlement (even if they think they are being nice about it) demanding changes, asking to support older versions of node, removing dependencies that annoy their build system, complain about how it's not working well with yarn, and other issues I don't care about.

NONE of these people USE hawk, or give a crap about the way I do things. I am not asking them to. These are developers that have not chosen to work within my ecosystem and my way of doing things. This was forced on them as much as on me.

It is also absurd that every month, hawk is downloaded 24 million times, hoek 26 millions times, boom 36 million times, cryptiles 24 million times, and sntp 24 million times FOR NO REASON! That's over 100 millions monthly download of code NO ONE uses.

I am always happy to try and solve real issues people have with my code. I take pride of my work and go out of my way to support people who honored me with building their product using it. But this is not part of the deal. You don't get to waste my time if you don't use my work in a meaningful way.

Please find the less disruptive way to remove this or make it optional in some way. For example, joi has a dependency on isemail that is structured in a way that the module is only loaded when needed which allows people to not require they don't want it.

At some point, I am going to take one-sided actions that will force this move. I am not going to break anything but I sure can annoy the fuck out of everyone with colorful console logs...

We are like a couple who finds out a year later that when got drunk one night and got married. Time for a quick annulment - we are already seeing other people.

hueniverse:
Closes #2831.

This replaces the hawk dependency with a local implementation that's identical.

This is the bare minimum the request community can do to disconnect this module from the hawk and hapi modules. The request community has been a major source of headache due to its inclusion of hawk and hoek. Because the maintainers have refused to remove this functionality, I am trying a simpler approach by moving the 100 lines of code from hawk here.

I will maintain this code if needed in the future which is better than you are getting today as I am no longer maintaining hawk v6 and this module cannot use v7 due to node requirements.


simov:
Sounds good to me. 

/cc @mikeal

simov:
I've published v2.87 with the patch.

PR #2973, opened 2018-07-06
Use correct tough-cookie-file-store in README.md

vishalkuo:
Following this [issue](https://github.com/mitsuru/tough-cookie-filestore/issues/3), it seems like `tough-cookie-filestore` has concurrency issues when saving to the file system resulting in incorrect JSON serialization [(see here)](https://github.com/Huachao/vscode-restclient/issues/31). 

I used this README as a reference on how to persist cookies and ran into this concurrency bug as I used `tough-cookie-filestore` instead of `tough-cookie-file-store`. I'm proposing this change to prevent future users looking for the same functionality from encountering the same issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1063,12 +1063,12 @@ request({url: url, jar: j}, function () {
 ```
 
 To use a custom cookie store (such as a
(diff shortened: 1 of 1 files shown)


kinkyell:
#2973 dfdfdfd

PR #2996, opened 2018-08-05
fix(uuid): import versioned uuid

kwonoj:
## PR Checklist:
       
## PR Description

This PR proposes way to import `uuid` module, per deprecation notice (https://github.com/kelektiv/node-uuid#uuid-) to not to import top level import directly. Currently top level import provides compatibility (https://github.com/kelektiv/node-uuid/blob/fe4ae79c55af2c7cbf2bb39d3bcb6716d5367091/index.js#L4) to forward to v4, but also it imports v1 still. 

This behavior is generally not a major concern, but few framework specifically like Electron (https://github.com/kelektiv/node-uuid/issues/189 / https://github.com/electron/electron/issues/2073 ) have issues with crypto performance, so v1's up front invocation `rng` costs amount of time on windows platform. 

Since top level import already points to v4, this change shouldn't cause any functional differences.


simov:
This seems to be fixed long time ago https://github.com/kelektiv/node-uuid/commit/501a8f4a3574b84a878f9b95c731a92d2f36ff40 so I don't see how changing the import now would make any difference.

simov:
Now I see the deprecation bits in the docs. Request does not call the `rng()` function upfront because `^` bumps to the latest minor.

Either way it's good to have the versioned import here :+1:

Issue #2954, opened 2018-05-25
FormData + Digest resulting in incorrect Content Length and 408: Client timeout


dbellavista:
### Summary

When using both `digest authentication` and `formData` the first request should be with `Content-Length: 0`. However the content-length is above 0 and web servers such as `nginx` will timeout the request with 408

I'm using request 2.87.0 on node 8.11

### Simplest Example to Reproduce

`ncat -l 8080 -k`

```js
request({
  method: 'POST',
  url: 'http://127.0.0.1:8080/test',
  auth: {
    sendImmediately: false,
    user: 'a',
    password: 'b'
  },
  formData: { a: 'a' }
},
```

### Expected Behavior

This headers are produced by `curl --http1.1 -F "a=a" http://127.0.0.1:8080/test --digest -u a:b`:

```
POST /test HTTP/1.1
Host: 127.0.0.1:8080
User-Agent: curl/7.54.0
Accept: */*
Content-Length: 0
```

And should be as expected.

### Current Behavior

Those headers are produced by the script:

```
POST /test HTTP/1.1
host: 127.0.0.1:8080
content-type: multipart/form-data; boundary=--------------------------277206183355889642026386
content-length: 157
Connection: close
```

As you can see the content length `157` is erroneous and web servers such as `nginx` configured to `proxy_pass` to the application server will timeout the request.

## EXTRA

By changing in the script `formData` to `json`, the content-length is still present (should be 0), however the body is passed:

```
POST /test HTTP/1.1
host: 127.0.0.1:8080
accept: application/json
content-type: application/json
content-length: 9
Connection: close

{"a":"a"}
```

For optimal performance, the `content-length` should be 0, however since the body is present, the request itself is correct.

### Your Environment

| software         | version
| ---------------- | -------
| request          | 2.87
| node             | 8.11.2
| npm              | 6
| Operating System | Ubuntu 16.04/MacOS

Issue #2777, opened 2017-09-22
HTTP Digest Authentication over POST


KingRial:
### Summary
I am 
unable to use Digest Authentication over a simple POST request

### Simplest Example to Reproduce
```javascript
request.post({
        url: 'URL to digest server',
        auth: {
          user: 'username',
          pass: 'password',
          sendImmediately: false
        },
        form: { lvalue: value }
      }, ( oError , oResponse , sBody) => {
... whatever you neeed ....
      });
```
### Expected Behavior
I expect to see two requests: the first should fail with error 401 and the second should be able to handle the original request

### Current Behavior
For some reasons I haven't understood the "sendImmediately" set to false is ignored.

When the request is launched, the method "basic" of "auth.js" is called twice; the first is using the "sendImmediately" option to "false"; the second one is set to "true".
The second internal call to the method seems to be the problem; it seems it doesn't receive the arguments set by the user.

### Possible Solution
I had to force "sendImmediately" to "false" before the following line: https://github.com/request/request/blob/master/lib/auth.js#L29 to see the library work correctly with HTTP Digetst

### Your Environment
| software         | version
| ---------------- | -------
| request          | ^2.81.0
| node             | 6.10.0
| npm              | 3.10.10
| Operating System | Windows 10

emilian:
I can't get digest authentication to work with PUT/POST either. Tried the method above but still not working. It seems like the PUT request is sent without the authorization header so it fails right away.

KingRial:
I am still using the previous workaround with POST methods.
Usually you also need to enable the "Cookies" to let the digest authentication work correctly

```
...
const request = require('request');
const cookies = request.jar();
request.defaults({
  jar: cookies
});
...
```

and then use the "jar" also in the request post configuration:

```
...
request.post({
        url: 'URL to digest server',
        auth: {
          user: 'username',
          pass: 'password',
          sendImmediately: false
        },
        jar: cookies,
        form: { lvalue: value }
      }, ( oError , oResponse , sBody) => {
... whatever you neeed ....
      });
...
```

emilian:
I still wasn't able to make it work with the request library, even after setting up the cookies. The authorization header just wasn't being sent at all. 

In the end, I went with a lower-level library that implements post from this link:

https://github.com/vinger4/node-http-digest-client

I only needed to upload a file to a WebDAV server using digest, so the library above worked for me.

Issue #28, opened 2011-05-05
Add body as a property to response?


aseemk:
Hey there,

How do you feel about adding `body` as a property to `response`? That would simplify callbacks to be just `function (error, response)`.

I've become a big fan of [Streamline.js](https://github.com/Sage/streamlinejs), which lets you write async code as if it was sync (really awesome <3). It makes the (admittedly not robust) assumption that async callbacks are of the form `(error, result)`, which lets you do stuff like:

```
try {
    var result = someAsyncTask(foo, bar, _);    // the _ is the callback placeholder
    // do something with result
} catch (error) {
    // do something with error
}
```

Which would otherwise be:

```
someAsyncTask(foo, bar, function (error, result) {
    if (error) {
        // do something with error
    } else {
        // do something with result
    }
});
```

The reason that's awesome is because this totally removes crazy indentation levels and makes code generally so much easier to follow and reason about (especially when it comes to error handling).

But of course, since JS as a language doesn't have a notion of multiple return values, any callback arguments beyond the first two are lost by Streamline. Hence my request. =)

In parallel, I'm working with the Streamline author to see if we can add a feature that automatically detects and "packs" multiple return values into an array:

```
var results = someAsyncTask(foo, bar, _);
// if someAsyncTask calls the callback w/ multiple values:
var a = results[0], b = results[1], c = results[3];
```

So the pattern for using request would look like:

```
var result = request.get(opts, _);
var response = result[0];
var body = result[1];
```

Which is reasonable. But I figure I'd ask -- would you be open to adding `body` as a property onto the `response` object to _streamline_ this? ;)

Thanks in advance!

PR #2998, opened 2018-08-07
fixed incorrect content-length on digest authentication request. (#28…

liefersfl:
…65 #2954 #2777)

## PR Checklist:
       
## PR Description

There was a bug in handling requests with two step authentication (eg. digest) and formdata. In the first step, the content-length of the formdata was calculated and sent in the request header as "content-length: xxx" but the content was not sent. The whole request ended in a timeout while the server was waiting for the content.

The following example produces a timeout:

 ```js
const options = {
    method: "POST",
    url: "<url to server>",
    auth: {
        user: 'abcde0ab963dc088cca337e19ebb6e95',
        pass: 'abcde1f223f4200e5b74273a1538ecc8752ddde93d6c8e45c50e55c9392a11ec',
        sendImmediately: false
      }
}

var r = request(options, function(err, httpResponse, body) {
    if (err) {
        console.log(err)
    }
    console.log("Done")
})

var form = r.form();
form.append('file', fs.createReadStream("/tmp/test.gz"), {
    filename: 'test.gz'
})
  ```

The fix sends a zero content-length when no content will be sent.

All tests passed, i just had to create new server certificates since they expired today ;-).


zz123er:
7 sep 2019  after so many years still this kind of bug   unbelievable  so sad!

Issue #2999, opened 2018-08-08
Does not handle multiple WWW-Authenticate headers


marcoancona:
### Summary
Check against `WWW-Authenticate` should take into account the possibility of receiving multiple headers

### Simplest Example to Reproduce
Send a request to an endpoint that returns multiple `WWW-Authenticate` headers, ie. `https://caldav.icloud.com`
```js
request({
  method: 'GET',
  url: 'https://caldav.icloud.com',
}, (err, res, body) => {
  console.log(res.headers['www-authenticate']) // "Basic realm=\"MMCalDav\", X-MobileMe-AuthToken realm=\"MMCalDav\""
}
```
Notice that, for this particular response, it would work as the first scheme is `Basic` and the second one is `X-MobileMe-AuthToken`. If it was the other way around, authentication would fail currently.

### Expected Behavior
Consider all `WWW-Authenticate` headers and proceed with one of the supported schemes, if any.

### Current Behavior
The library does not recognize that one of the supported authentication schemes is available and the request fails (401).

### Possible Solution
Improve lines 151-152 of https://github.com/request/request/blob/925849a8278ae794eb9c257cf4d6cc4fb3ee89a8/lib/auth.js

### Context
Authenticate against CalDAV servers.

### Your Environment
| software         | version
| ---------------- | -------
| request          | 2.87.0

marcoancona:
Afaik this is still an open issue

PR #3000, opened 2018-08-08
Support for multiple www-authenticate headers

marcoancona:
https://github.com/request/request/issues/2999

## PR Description
Adding support for multiple `www-authenticate` headers. Parsing of `www-authenticate` from `response.headers` requires some care, as multiple headers are joined using `,` which is also used inside the `Digest` header. So we use a regex to split the string, looking for a capital letter following the comma.


marcoancona:
Mmmm... https tests have an issue (CERT_HAS_EXPIRED). This also the same on `master` for me.

PR #3005, opened 2018-08-10
test: moving to tap for test runner on nodejs.

mikeal:
The old runner, taper, is not maintained and was causing lots of issues. Migrating directly to tape caused all kinds of other issues. In this migration to tap I noticed new timing issues for flaky tests and even some tests that weren't being run by taper.


mikeal:
I also fixed the TLS certs in this push as well.

Issue #2327, opened 2016-08-12
Malformed host header when accessing a docker daemon


manast:
This may not be a bug in request, and I would be happy to report it to the docker team if that is the case. The thing is that after upgrading to the latest version of docker (1.12), when I tried to connect to the daemon using unix domain sockets, I got the error:
400 Bad Request: malformed Host header'

I noticed that the host header sent by request was: `/var/run/docker.sock`

On the other hand, using curl it worked well, and the request looked like this:

```
$ curl -v --unix-socket /var/run/docker.sock http:/version
*   Trying /var/run/docker.sock...
* Connected to http (/var/run/docker.sock) port 80 (#0)
> GET /version HTTP/1.1
> Host: http
> User-Agent: curl/7.47.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Type: application/json
< Server: Docker/1.12.0 (linux)
< Date: Fri, 12 Aug 2016 13:14:44 GMT
< Content-Length: 198
```

And by overriding the host header, I also got request working:

``` js
request({url: 'http://unix:/var/run/docker.sock:/version', headers:{host: 'http'}}, function(err, response, body){...})
```

Could it be that request is not setting the host header correctly?

mindnuts:
@manast I love you man (no homo)...i was facing the same issue and infact i opened a ticket on [Docker] (https://github.com/docker/docker/issues/26099)

Seems like the underlying implementation changed and it stopped working for me since Docker 1.12

mindnuts:
@manast  BTW...to connect using Curl now you would have to use:
```bash
curl -v --unix-socket /var/run/docker.sock http://localhost/containers/json
```

dfrankland:
If this helps anyone, I was able to get this to work with the built-in `http` module like below:
```js
const http = require('http');

const options = {
  socketPath: '/var/run/docker.sock',
  path: '/containers/json',
};

const callback = res => {
  console.log(`STATUS: ${res.statusCode}`);
  res.setEncoding('utf8');
  res.on('data', data => console.log(data));
  res.on('error', data => console.error(data));
};

const clientRequest = http.request(options, callback);
clientRequest.end();
```

thonatos:
@dfrankland thanks. 

```
STATUS: 200
[ { Containers: -1,
    Created: 1483565663,
    Id: 'sha256:53df695896acdecb39a6914e25f7680011be26504dcb0675ac6e98c49379d171',
    Labels: {},
    ParentId: '',
[...]

PR #3035, opened 2018-10-01
fix: #2327, Host header is not required in UNIX Domain Sockets

lxs137:
## PR Checklist:
       
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: [#2327]
       

## PR Description

If add Host header to request through UNIX Domain Sockets, some server(Docker Engine API Server) may reject it, it is also mentioned in #2327.
Host header must be sent in HTTP/1.1, but if we use UNIX Domain Sockets, it is not required to automatically add it to the request headers.

### DIFF
--- a/tests/test-headers.js
+++ b/tests/test-headers.js
@@ -263,6 +263,31 @@ tape('catch invalid characters error - POST', function (t) {
   })
 })
 
+tape('Host header is not required when use UNIX Domain Sockets', function (t) {
+  if (os.platform() === 'win32') {
+    t.end()
+  }
+  var sockPath = '/tmp/request.sock'
+  var sockServer = server.createUnixSockServer(sockPath)
+  sockServer.on('/', function (req, res) {
+    res.writeHead(200, {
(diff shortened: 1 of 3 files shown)


achingbrain:
As a workaround you can also specify the host header as an empty string

```javascript
request({
  uri: 'http://unix:/path/to/socket..etc',
  headers: {
    host: ' '
  }
}, ...)
```

tests/test-headers.js
```diff
@@ -263,6 +263,31 @@ tape('catch invalid characters error - POST', function (t) {
   })
 })
 
+tape('Host header is not required when use UNIX Domain Sockets', function (t) {
+  if (os.platform() === 'win32') {
+    t.end()
```
reconbot: Probably want to return

lxs137:
I find some apps (e.g. MySQL) that use the approach of "listen on localhost" means "use the UNIX Domain socket" instead of TCP/IP. For these apps, I think they can process the http header over UNIX Domain socket.

PR #3053, opened 2018-10-30
Support a runImmediate flag to avoid processing delays

turt2live:
## PR Checklist:
       
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A
       

## PR Description
This is particularly useful in web environments where setImmediate() may be temporarily paused by the browser. When a high resource usage tab is put in the background, browsers often pause timer usage until the app is foregrounded again. For some web applications, performing background network requests even when backgrounded itself is useful and may wish to avoid use of setImmediate, understanding the risk in potentially doing so.


turt2live:
Looks like the test failures are unrelated and also present on `master`.

PR #3084, opened 2018-12-10
Allow content-length of 0 (#920)

### LINKED ISSUE #920: Sending requests with content-length equal to zero raises an error
In the request.js file there is a check:
if (length) {
    if (!self.hasHeader('content-length')) self.setHeader('content-length', length)
} else {
    throw new Error('Argument error, options.body.')
}
which prevents sending files with zero length. I understand this not very important issue, but still. Specification says "Any Content-Length greater than or equal to zero is a valid value." (http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html)

I faced with this restriction when using cradle which we use to save user files to CouchDB and some of files have zero length (CouchDB also accepts empty files), so the problem is caused only by the request module.

pretorh:
## PR Checklist:
       
- [ ] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A]
       

## PR Description
Change check for if length is set in order to allow passing a 0-length buffer as body. Previously this emited an error. This is to fix #920

### DIFF
--- a/request.js
+++ b/request.js
@@ -431,7 +431,7 @@ Request.prototype.init = function (options) {
         length = self.body.length
       }
 
-      if (length) {
+      if (length !== undefined) {
         self.setHeader('content-length', length)
       } else {
         self.emit('error', new Error('Argument error, options.body.'))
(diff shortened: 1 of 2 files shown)


stephenh:
FWIW we ran into this as well, would be great to have this merged. Thansk @pretorh !

jdmarshall:
Also running into this, and agree with the proposed solution.

PR #3094, opened 2019-01-08
Use tough-cookie 3.x

stash:
## PR Checklist:
- [ ] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A

## PR Description
Upgrade `tough-cookie` to the 3.x API. Since `request` doesn't need to use the new API, nor is it affected by `tough-cookie` updated node engine requirement (>= 6), there are therefore no code or test changes required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,7 +679,7 @@ request.get({
 
 The `ca` value can be an array of certificates, in the event you have a private or internal corporate public-key infrastructure hierarchy. For example, if you want to connect to https://api.some-server.com which presents a key chain consisting of:
 1. its own public key, which is signed by:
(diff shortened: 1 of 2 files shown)


mikeal:
Why is CI failing?

stash:
Looks like CI was probably failing because of https://github.com/salesforce/tough-cookie/issues/140 - basically, tough-cookie depended on an ip-regex that was too new for that node engine version. Resubmitted with 3.0.1 which has a more compatible ip-regex

stash:
OK nope, that wasn't it... at least I have TravisCI running on my fork now. Continuing to investigate.

Issue #3093, opened 2019-01-05
Digest authentication fails against MusicBrainz API


Borewit:
### Summary
Digest authentication against [MusicBrainz API](https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2#Authentication) fails.

Related and maybe the same cause: #3072

### Simplest Example to Reproduce

```JavaScript
/**
 * Related API doc: https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2#ISRC_submission
 */
public async _request_post(entity: 'recording', xmlMetadata: XmlMetadata): Promise<any> {

    return new Promise<void>((resolve, reject) => {

      request.post({
        url: `'http://musicbrainz.org/ws/2/${entity}/`, // Use HTTPS for production!!!
        body: xmlMetadata,
        headers: {
          'Content-Type': 'application/xml',
          'User-Agent': user_agent
        },
        qs: {
          // fmt: 'json', // This seems to confuse the MusicBrainz API
          client: 'WhatMusic/0.0.1'
        },
        // proxy: 'http://localhost:8888', // Redirect to Fiddler
      }, (error, response) =>  {
        if (error) {
          reject(error);
        } else {
          if (response.statusCode === HttpStatus.OK) {
            resolve(response.body);
          } else {
            reject( new Error('MusicBrainz API error'));
          }
        }
      }).auth(this.config.username, this.config.password, false)

    });
  }
```

### Expected Behavior
Expect the digest authentication to work.

### Current Behavior
Receive a 401 (UNAUTHORIZED) from MusicBrainz, with following message:

> Your credentials could not be verified. Either you supplied the wrong credentials (e.g., bad password), or your client doesn't understand how to supply the credentials required.

Header generated by request:
```
authorization: Digest username="Borewit", realm="musicbrainz.org", nonce="6CCUtvEQ6RGDfthcSlGD+w==", uri="http://musicbrainz.org/ws/2/recording/?client=WhatMusic-0.0.3", qop=auth, response="1eb1140de19a680210127f847a7803a7", nc=00000001, cnonce="0994fcf97892413987ed7c83f8a2e131", algorithm=MD5, opaque="QCiUtvEQ6RGDfthcSlGD+w=="
```
### Possible Solution
I compared the differences with a successful CURL using Fiddler, and noticed the `uri` field in the HTTP 

For comparison I tried to to the same thing with CURL:
``` bash
[...]

PR #3101, opened 2019-02-04
Use relative uri path to fix digest authentication

Borewit:
Fix issue with absolute URL instead of relative URL passed to digest authentication.
Issue: #3093


Borewit:
Please keep open.

PR #3112, opened 2019-02-19
Capitalize Authorization header

jspahrsummers:
## PR Checklist:
- [x] I have added/updated tests for any new behavior. (Nothing worth testing here, beyond the existing suite.)

## PR Description
This is a dumb change! Although HTTP headers are supposed to be case-insensitive, not all servers behave that way in practice—I found a random API that wouldn't behave unless `Authorization` is capitalized when using HTTP basic auth.


jspahrsummers:
Oops, this is #3041.

PR #3113, opened 2019-02-21
Stop wrapping cookie method in request.defaults

pro-src:
For some odd reason when [cookie support](https://github.com/request/request/pull/102) was first added in 2011, the `request.cookie` method was treated like a request method. All it does is parse a cookie string. Everybody has been wrapping it since then... lmao

I'd guess the reason why this has gone undiscovered is because nobody has ever tried to do the following.
```js
const request = require('request').defaults({});
// This will make a request!!!
request.cookie('some cookie string');
```

Anyway, this PR fixes that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -118,7 +118,7 @@ request.defaults = function (options, requester) {
     defaults[verb] = wrapRequestMethod(self[verb], options, requester, verb)
   })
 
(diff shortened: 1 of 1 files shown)


pro-src:
I'd like to propose other changes to index.js once somebody has the time to review this PR.

PR #3115, opened 2019-02-25
Update README timeout

jaredsuttles:
## PR Checklist:
       
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A]
       

## PR Description
Timeout argument description is currently very misleading. It was split into [two bullets here](https://github.com/request/request/pull/2728) but the note regarding OS-wide timeout does not apply to the first bullet, but rather the second. In addition, [this change](https://github.com/request/request/pull/1392/files) made the wording more confusing, in my opinion. I reverted it back to what it was originally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -823,12 +823,9 @@ The first argument can be either a `url` or an `options` object. The only requir
     work around this, either use [`request.defaults`](#requestdefaultsoptions)
     with your pool options or create the pool object with the `maxSockets`
     property outside of the loop.
-- `timeout` - integer containing number of milliseconds, controls two timeouts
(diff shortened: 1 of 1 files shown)


lalau:
I am still confused with the timeout doc. Does the value apply to both bullet points (read timeout and connection timeout)? Or can it set different value to them?

jaredsuttles:
@lalau "integer containing number of milliseconds, controls two timeouts." then those two timeouts are listed below. So yes, it applies to both. I think it makes sense now but maybe you have a suggestion for further clarity?

PR #3149, opened 2019-04-10
Added a maxResponseSize option that limits the maximum buffered response size

mivanovaxway:
Added maxResponseSize option, checks for invalid values, backward compatibility and relevant tests plus fix to the test-timeout.js so all the tests pass

## PR Checklist:
- [Y] Relevant documentation

- [Y] Full backward compatibility is ensured

## PR Description
When using a callback, the response is being buffered either in Buffer or string array.
The size of this buffer however is not regulated and there is no way to explicitly limit it.
The same also applied when the response is compressed, after decompression it can become much bigger leading to out-of-memory problems.

This feature adds an optional limit(maxResponseSize) that can protect you from such issues. 
Please see the updated Doc for more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -805,7 +805,7 @@ The first argument can be either a `url` or an `options` object. The only requir
 - `removeRefererHeader` - removes the referer header when a redirect happens (default: `false`). **Note:** if true, referer header set in the initial request is preserved during redirect chain.
 
 ---
-
+<a id="request-encoding-section"></a>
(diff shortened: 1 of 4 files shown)


jasondecastro:
I'm happy I stumbled across this, was pretty helpful to my use-case. Thanks! 

Hoping it gets merged in - it's a great option to have. Prior to this, I found that reading the length of the Buffer after-the-fact often led to some "Out of memory" heap issues if the buffer was too large. This seems to calculate it by the chunk though, which solves that problem.

thibauds:
Thank you @mivanovaxway, I would be very happy to see it merged!

PR #3166, opened 2019-05-21
Adds logger option

yuric18:
## PR Checklist:

## PR Description
Adds a `logger` option so that all requests are logged at `info` level. 
It supports the standard logging API (like `console` or Winston).

### How it's done:
- If `logger` option is provided, `self.timings` is set to `true` so it can use `self.logger.info` using a few variables so its output is like below:
`----> POST http://localhost:50240/ - 200 - 22.95ms` (retrieved from test).
- The arrow is to separate the logs that the server did from logs that the server received, so it don't get confusing. (Maybe there is another way to do this? Yeah, of course there is, but that's what I was able to think).

### Reason for this PR
- `options.time` works, but you have to manipulate the timings to log the requests you made.
- I've asked myself "why not to make this a built-in feature of the package?". Seems reasonable to me :smiley:.

### DIFF
--- a/README.md
+++ b/README.md
@@ -865,6 +865,7 @@ The first argument can be either a `url` or an `options` object. The only requir
     - `firstByte`: Duration of HTTP server response (`timings.response` - `timings.connect`)
     - `download`: Duration of HTTP download (`timings.end` - `timings.response`)
     - `total`: Duration entire HTTP round-trip (`timings.end`)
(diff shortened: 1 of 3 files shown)


ecasilla:
Is this PR still being worked on? Its very useful to allow logger swapping

yuric18:
@ecasilla actually its already done. I'm only waiting for the PR approval to be merged. For now, I'm using my fork as a project dependency to use this feature while it's not merged.

PR #3200, opened 2019-08-31
Mitigate prototype pollution when filtering options

snyamathi:
## PR Checklist:
- [x] I have run `npm test` locally and all tests are passing. ( `test-localAddress.js` fails in my branch, but also on `master` so I assume something with my env differs from the ci-build where it may pass )
       
- [ ] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A

## PR Description
In an ideal world, prototype pollution wouldn't be an issue and we could prevent vectors that ourselves or our colleagues may open.

Unfortunately, this is impossible to guarantee.  While we can do our best to mitigate the risk of prototype pollution by passing a clean `options` object to `request`, the method `filterForNonReserved` creates a new object with the possibly polluted `object` prototype rather than either the clean object we have provided or filtering the object in place.

This PR is very subtle - the only change is from

```js
var object = {}
```

to

```js
var object = Object.create(null)
```

Though for those of us that pass in a clean object, we won't get a polluted object out.  If someone is using a polluted prototype intentionally, those options will be passed since we're not filtering the `for ... in` loop with `Object.hasOwnProperty`.

Practically speaking, if our node process's Object's prototype has a polluted value for `host` we will have `request` make requests to potentially malicious hosts.

Current attack vector:
```js
// Contrived example, but this could come from very sneaky places
var obj = {}
obj.__proto__.hostname = 'yahoo.com'

// Create a clean object, not affected by previous prototype pollution
var options = Object.create(null);
options.uri = 'http://localhost'
console.log(options.hostname) // undefined

// This will _still_ hit our polluted hostname!
request(options, (err, res, body) => console.log(body))
```

### DIFF
--- a/lib/filterForNonReserved.js
+++ b/lib/filterForNonReserved.js
@@ -1,0 +1,15 @@
+function filterForNonReserved (reserved, options) {
+  // Filter out properties that are not reserved.
+  // Reserved values are passed in at call site.
+
+  var object = Object.create(null)
+  for (var i in options) {
+    var notReserved = (reserved.indexOf(i) === -1)
+    if (notReserved) {
+      object[i] = options[i]
+    }
+  }
+  return object
+}
+
+module.exports = filterForNonReserved
--- a/request.js
+++ b/request.js
@@ -18,6 +18,7 @@ var isstream = require('isstream')
 var isTypedArray = require('is-typedarray').strict
 var helpers = require('./lib/helpers')
 var cookies = require('./lib/cookies')
+var filterForNonReserved = require('./lib/filterForNonReserved')
 var getProxyFromURI = require('./lib/getProxyFromURI')
 var Querystring = require('./lib/querystring').Querystring
 var Har = require('./lib/har').Har
@@ -30,7 +31,6 @@ var Tunnel = require('./lib/tunnel').Tunnel
 var now = require('performance-now')
 var Buffer = require('safe-buffer').Buffer
 
-var filterForNonReserved = helpers.filterForNonReserved
 var safeStringify = helpers.safeStringify
 var isReadStream = helpers.isReadStream
 var toBase64 = helpers.toBase64
--- a/request.js
+++ b/request.js
@@ -30,6 +30,7 @@ var Tunnel = require('./lib/tunnel').Tunnel
 var now = require('performance-now')
 var Buffer = require('safe-buffer').Buffer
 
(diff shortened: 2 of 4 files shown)


request.js
```diff
@@ -30,6 +30,7 @@ var Tunnel = require('./lib/tunnel').Tunnel
 var now = require('performance-now')
 var Buffer = require('safe-buffer').Buffer
 
+var filterForNonReserved = helpers.filterForNonReserved
```
snyamathi: I've moved this function into `helpers` to make clear the attack vector and this change's mitigation via unit tests.  If there is a better or different way to organize this, I'm all ears.

tests/test-filterForNonReserved.js
```diff
@@ -0,0 +1,27 @@
...
+  obj.__proto__.hostname = 'evil.com' // eslint-disable-line no-proto
+  t.end()
+})
+
+tape('do not create a polluted object when filtering reserved options', function (t) {
+  var options = Object.create(null)
+  assert.equal(options.hostname, undefined)
+
+  var filtered = filterForNonReserved([], options)
+  assert.equal(filtered.hostname, undefined)
```
snyamathi: Previously even though we're passing in a clean object, the resulting object from the filter method would be dirty.

snyamathi:
Since the coverage decreased, we must not have tests for the helper functions, I'll move the `filterForNonReserved` to its own file.

lib/filterForNonReserved.js
```diff
@@ -0,0 +1,15 @@
+function filterForNonReserved (reserved, options) {
+  // Filter out properties that are not reserved.
+  // Reserved values are passed in at call site.
+
+  var object = Object.create(null)
+  for (var i in options) {
```
lingyan: I wish there is a `if (options.hasOwnProperty(i))` check here too.   The tricky thing about adding that check is it might break existing code using it.

stevengpwc:
Hi maintainers @lingyan @snyamathi , when will this fix be released to the next version?

snyamathi:
@stevengpwc  I believe this repo is maintenance only at this point and I don't expect this PR to be merged or this issue to be fixed.  On our end, we've internally made some checks for prototype pollution and I would suggest that you do the same.  Good luck!

richaagr:
@mikeal Is there a plan to merge this change soon? Its causing our vulnerability score to go down.

PR #3211, opened 2019-10-08
tough-cookie-version-upgrade

zeeshanpaalo:
## PR Checklist:
       

fbray:
Also looking to upgrade tough-cookie version to a stable version.  Should we log an issue in github here?  Or is it possible to get this PR passing tests?

Issue #3365, opened 2021-02-19
ERR_STRING_TOO_LONG on large file downloads


vrsbrazil:
https://github.com/request/request/blob/3c0cddc7c8eb60b470e9519da85896ed7ee0081e/request.js#L1128

This line throws ERR_STRING_TOO_LONG  error on large downloads, i just ignored this line hardcoding a false on line 1127 clause to get my work done, but just sending you guys this message so you can take a look.

Thank you.

makingglitches:
Trying to debug this.
The stacktrace in my code indicates the error is being emitted from the following code in request.js:

At line 1128.

    if (self.encoding !== null) {
        response.body = response.body.toString(self.encoding)
      }

is where the problem seems to be emanating.

Core of the issue seems limits in size eg.

Maximum javascript string length is 256 MB
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length

Buffer can hold up to 1 GB

Normally I note that the pipe() argument is sending data as available through the write stream, that's what it is supposed to do. How is this happening I do not know.


So a trace through of the following code setup:

> var req = request.get(url);
> // some setup code
> req.pipe(ostream);
> 

the initial function call that sets up the request hits a generic wrapper verbFunc at line 66 of the code.
which returns a type from line 42  both of index.js which resolves to 
request of request.js at line 93.

then it hits pipe at request.js line 1467 and what is actually occurring becomes extremely unclear.
there appears to be asynchronous calls that don't seem to be started anywhere occuring.

I'm trying to do a traceback to find the logic in question.

The error is happening in readResponseBody line 1198.
which is being called at onRequestResponse  line 880
which is being registered at line 775
which is part of Request.start at line 708
which gets called as part of a error of connection reset (not relevant) 

at 1491  of  Request.write at 1486
at 1505 of Request.end  at 1497

I placed breakpoints at  all instances of Request.write to see where its hitting this.

I know how i'd fix the problem, I'd put a cap on buffer size that gets sent to the stream and defer further output for later to fix the problem with the differing size allowances.


==> updates <==


the request.end seems to be called when not using pipe to finish writing and called prior to the job start in the case of pipe.
none of the breakpoints for request.write are getting hit.

a simplified debugging scenario for deeper trace is executing without a hitch as the ONLY DOWNLOAD.
[...]

PR #3378, opened 2021-05-10
Removing lines 1127 to 1129

vrsbrazil:
## PR Checklist:
- [X ] I have run `npm test` locally and all tests are passing.
       
- [ X] If this is a significant change, an issue has already been created where the problem / solution was discussed: [ https://github.com/request/request/issues/3365 ]

## PR Description
After analysis on issue #3365 was not able to find a direct use for these lines of code and it actually would raise ERR_STRING_TOO_LONG on large file downloads.


mrkmg:
After a quick glance, this would be a breaking change if one was to expect response.body to be a string but actually received a buffer.

PR #3425, opened 2022-08-29
-- upgrade dependencies

pavhov:
## PR Checklist:
       

UserAdminD3us (approved):

CHANGELOG.md
```diff
@@ -443,7 +443,7 @@
 - [#1116](https://github.com/request/request/pull/1116) Fix typo in `noProxyHost` definition (@FredKSchott)
 - [#1114](https://github.com/request/request/pull/1114) linting fix: Added a `new` operator that was missing when creating and throwing a new error (@FredKSchott)
 - [#1096](https://github.com/request/request/pull/1096) No_proxy support (@samcday)
-- [#1107](https://github.com/request/request/pull/1107) linting-fix: remove unused variables (@FredKSchott)
+- [#1107](https://github.com/request/request/pull/1107) linting-fix: remove unused letiables (@FredKSchott)
```
rtharston, line 446: It looks like you performed a search replace on var with let, and it went a bit too far afield.
There are quite a few of these in the changelog and readme, and probably other places too, they'll need changed back.

index.js
```diff
@@ -125,7 +125,7 @@ request.defaults = function (options, requester) {
 }
 
 request.forever = function (agentOptions, optionsArg) {
-  var options = {}
+  const options = {}
```
rtharston, line 128: What do all these var to const replacements have to do with upgrading dependencies?

Did you run a linter or something?

rtharston (changes requested):
The title of the PR says you upgraded the dependencies, but this includes a lot of other changes. Please only update the dependencies and revert the rest.
Updating var to const where possible isn't a bad thing, but is not within the scope of your stated change.

PR #3473, opened 2023-11-27
log buffers over limit

BrendanMeintjes:
## PR Checklist:
       
### DIFF
--- a/request.js
+++ b/request.js
@@ -1,96 +1,96 @@
-'use strict'
+"use strict";
 
-var http = require('http')
-var https = require('https')
-var url = require('url')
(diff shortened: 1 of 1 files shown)


request.js
```diff
@@ -1,806 +1,865 @@
...
-function filterForNonReserved (reserved, options) {
+"use strict";
+
+var http = require("http");
+var https = require("https");
+var url = require("url");
+var util = require("util");
+var stream = require("stream");
+var zlib = require("zlib");
+var aws2 = require("aws-sign2");
```
sridhar5969, line 9: changed only ' to "

